PR #35, opened 2014-07-20
Force cURL to use ipv4 for Dns resolving

elfif:
I had an issue today trying rollbar for the first time. After a few debugs i realised that curl_exec returned false.
curl_error returned it "cannot resolve host". After a few searches i managed to make it works by adding that curl option that forces it to resolve hostname into an IPV4 address.

I tested it on a vagrant VM using virtualBox on Windows 8.1
My VM : Ubuntu server 14.04, PHP 5.5.14, cUrl 7.35.0


Bilge:
This definitely should not have been accepted 😂

Issue #41, opened 2015-01-08
json_encode: invalid utf-8 sequence in argument (in $_GET/$_POST keys)


rfink:
Similar issue to the $_GET/$_POST *values*, but now it's on the keys.  Can be seen with common vulnerability scanners.  Example:

http://{yourdomainname.com}/{somepath}?acux8277%C0%BEz1%C0%BCz2a%90bcxuca8277

With:
display_errors Off (this is the only condition that causes the json_encode warning to happen)

Pull request coming shortly

dlsteuer:
This has been completed and released.

PR #42, opened 2015-01-08
Bugfix/jsonencodewarning

rfink:
I've tested this to the best of my ability, but since there aren't unit tests available, I'm only as certain as my limited testing can allow.

Edit:  fixes issue #41


brianr:
@rfink - thanks for this! Sorry it took so long to get merged. This is out now in version 0.10.0.

PR #44, opened 2015-02-12
Add proxy option

Laykou:
Add proxy option to the rollbar configuration. Can be set during initialization:

```PHP
// With authentication
$config = [
  'proxy' => [
    'address' => '123.123.123.123:9876',
    'username' => 'my_name',
    'password' => 'my_password'
  ]
];

// Or you can use only the address:
$config = [
  'proxy' => '123.123.123.123:9876'
];

Rollbar::init($config);
```


coryvirok:
Thanks @Laykou!

PR #46, opened 2015-03-03
Fix php < 5.4 compatibility

unix1:
The proxy feature merged in #44 introduced the array syntax that only works on php versions >=5.4 per http://php.net/manual/en/language.types.array.php introducing an implicit version dependency. This changes the syntax to `array()` so it still works on older versions of php.


brianr:
Thanks @unix1 !

Laykou:
Yeah, sorry guys, my bad. I work with version 5.5 now. @unix1  thanks :)

PR #47, opened 2015-03-29
Added custom & payload access to report_exception

veloper:


brianr:
Thanks @veloper ! This is merged and in version 0.10.0.

One note -- I did tweak the location of where 'extra' data goes, to be more consistent with our other notifiers. It's now inside `body.trace.extra`.

PR #49, opened 2015-03-31
Added trace chain of previous exceptions.

lindyhopchris:
Added logging of previous exceptions, which is support by the Rollbar API.

Can this be merged as soon as possible? At the moment the exceptions that we are logging to Rollbar are effectively useless, because they do not show the previous exceptions. In most PHP Frameworks (we're using Zend 2) the framework will catch and re-throw original exceptions, so we're not currently able to track down the original bug without previous exceptions also being logged.

### DIFF
--- a/src/rollbar.php
+++ b/src/rollbar.php
@@ -176,6 +176,10 @@ class RollbarNotifier {
 
     public function report_exception($exc, $extra_data = null, $payload_data = null) {
         try {
+            if (!$exc instanceof Exception) {
+                throw new Exception('Report exception requires an instance of Exception.');
+            }
+
             return $this->_report_exception($exc, $extra_data, $payload_data);
         } catch (Exception $e) {
             try {
@@ -232,9 +236,9 @@ class RollbarNotifier {
     }
 
     /**
-     * @param $exc Exception
+     * @param Exception $exc
      */
-    protected function _report_exception($exc, $extra_data = null, $payload_data = null) {
+    protected function _report_exception(Exception $exc, $extra_data = null, $payload_data = null) {
         if (!$this->check_config()) {
             return;
         }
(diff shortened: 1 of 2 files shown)


brianr:
Thanks @lindyhopchris ! We'll review shortly.

src/rollbar.php
```diff
@@ -176,6 +176,10 @@ public function __construct($config) {
 
     public function report_exception($exc, $extra_data = null, $payload_data = null) {
         try {
+            if (!$exc instanceof Exception) {
+                throw new Exception('Report exception requires an instance of Exception.');
```
brianr: I think this is probably the right thing to do, but it is a backwards-incompatible change. Will bump the version number to 0.11.0.

src/rollbar.php
```diff
@@ -246,24 +250,12 @@ protected function _report_exception($exc, $extra_data = null, $payload_data = n
...
-                    'message' => $message
-                )
-            )
+            'trace' => $this->build_exception_trace($exc, $extra_data),
         );
 
-        if ($extra_data !== null) {
-            $data['body']['trace']['extra'] = $extra_data;
+        if ($exc->getPrevious() instanceof Exception) {
+            $data['body']['trace_chain'] = $this->build_exception_trace_chain($exc);
```
brianr: The Rollbar API expects there to be either a 'trace' OR a 'trace_chain', but not both. I'll push a commit to fix this.

tests/RollbarNotifierTest.php
```diff
@@ -224,7 +224,7 @@ public function testExceptionWithExtraAndPayloadData() {
         try {
             throw new Exception("test");
         } catch (Exception $e) {
-            $uuid = $notifier->report_exception($uuid, array('this_is' => 'extra'), array('title' => 'custom title'));
+            $uuid = $notifier->report_exception($e, array('this_is' => 'extra'), array('title' => 'custom title'));
```
brianr: Nice, thanks

brianr:
Thanks @lindyhopchris ! This is now pushed in version 0.11.0.

lindyhopchris:
Excellent, thanks!

PR #50, opened 2015-04-02
Current URL Detection: Use port 443 if protocol is https and no port is detected

digilist:
My webserver configuration doesn't publish the SERVER_PORT (only god knows why...). So if the protocol is https, do not use the default port 80, but instead use 443.

### DIFF
--- a/src/rollbar.php
+++ b/src/rollbar.php
@@ -546,6 +546,8 @@ class RollbarNotifier {
             $port = $_SERVER['HTTP_X_FORWARDED_PORT'];
         } else if (!empty($_SERVER['SERVER_PORT'])) {
             $port = $_SERVER['SERVER_PORT'];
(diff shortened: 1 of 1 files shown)


src/rollbar.php
```diff
@@ -546,6 +546,8 @@ protected function current_url() {
             $port = $_SERVER['HTTP_X_FORWARDED_PORT'];
         } else if (!empty($_SERVER['SERVER_PORT'])) {
             $port = $_SERVER['SERVER_PORT'];
+        } else if ($proto == 'https') {
```
brianr: Think we should use `===` here. I'll make this change after merging.

brianr:
Merged, thanks! We'll get a new release out shortly.

brianr:
Released in version 0.15.0

PR #54, opened 2015-05-16
Require the curl extension in composer.json

jenssegers:
I would also add a minimum PHP requirement.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,9 @@
   "autoload": {
     "classmap": ["src/rollbar.php"]
   },
+  "require": {
+    "ext-curl": "*"
+  },
   "require-dev": {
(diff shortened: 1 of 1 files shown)


coryvirok:
Thanks for the PR @jenssegers. PHP isn't my primary language so can you help me understand the need and any potential downsides for this change?

Thanks!

jenssegers:
This will warn users that try to install rollbar when they do not have the curl extension installed.

jenssegers:
I did the same PR on Sentry's client: https://github.com/getsentry/raven-php/commit/6faf446d04c66809a4a08502eb2bd352c1620bff

coryvirok:
Gotcha. Thanks for the info.

brianr:
Thanks @jenssegers !

PR #56, opened 2015-05-27
[Fix] Add new error type E_PARSE(number:4)

diogopms:
Resolve the problem of not sending syntax errors to rollbar.
Please check if everything is ok.
Thanks


brianr:
Merged, thanks!

PR #57, opened 2015-06-04
Improve param scrubbing

chanind:
This PR adds several improvements to param scrubbing:

- regexes can be used in the `scrub_fields` list. ex: `$config['scrub_fields'] = array('/password/i');`
- If a param has an array as a value, the array will be recursively scrubbed
- `$_GET` and headers are also scrubbed before being recorded

### DIFF
--- a/src/rollbar.php
+++ b/src/rollbar.php
@@ -464,7 +464,7 @@ class RollbarNotifier {
             );
 
             if ($_GET) {
-                $request['GET'] = $_GET;
+                $request['GET'] = $this->scrub_request_params($_GET);
             }
             if ($_POST) {
                 $request['POST'] = $this->scrub_request_params($_POST);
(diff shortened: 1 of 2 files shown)


brianr:
Wow, thanks! We'll review and get this merged shortly.

src/rollbar.php
```diff
@@ -480,10 +480,15 @@ protected function build_request_data() {
...
+        $potential_regex_filters = array_filter($this->scrub_fields, function($field) {
+            return strpos($field, '/') === 0;
+        });
         foreach ($params as $k => $v) {
-            if (in_array($k, $this->scrub_fields)) {
-                $count = is_array($v) ? count($v) : strlen($v);
-                $scrubbed[$k] = str_repeat('*', $count);
+            if ($this->_key_should_be_scrubbed($k, $potential_regex_filters)) {
+                $scrubbed[$k] = $this->_scrub($v);
+            } elseif (is_array($v)) {
```
brianr: If I'm reading this correctly, the behavior here is:

- If the key matches a scrub pattern, then scrub it, no matter what its value is (i.e. even if the value is an array)
- Otherwise, if it's an array, recursively scrub params in that array

I think that sounds like the right thing to do, but I just thought I'd shine some light. Perhaps there should be a test for this?

chanind: yeah, that's intended. test has been amended here: cdcb2f147d7882bd7cf6683af713d1b77ca75b8e

brianr:
This is up now on packagist as version 0.12.0. Thanks again!

chanind:
awesome thanks for the fast turnaround on this!

PR #59, opened 2015-06-11
scrubbing URL query params

chanind:
This PR extends scrubbing to the URL that shows up in rollbar. Currently if a sensitive param is passed as a URL query parameter, it's scrubbed in the GET params list, but shows up in plain text as part of the URL.


brianr:
Looks good to me. Thanks!

brianr:
Released in 0.13.0

PR #60, opened 2015-06-22
call mt_srand() to guarantee fresh seed before generating a uuid

jontsai:


brianr:
Thanks @jontsai !

For anyone else following along: the bug that @jontsai ran into is that if your application happens to call `mt_srand` with a predictable value, and then later call a Rollbar report method, the generated UUIDs will also be predictable. Since Rollbar drops events with UUIDs that have already been seen (to avoid processing duplicates), that means that this situation will cause Rollbar to drop desired data.

brianr:
Released as 0.14.0

PR #63, opened 2015-07-08
Make scrub key fields comparing case insensitive

dimkalinux:


brianr:
Thanks @dimkalinux ! We'll get a new release out shortly.

PR #64, opened 2015-07-22
Preventing integer 0 keys from getting wrongly scrubbed

violuke:


brianr:
Thanks @violuke !

Can you tell me a little more about the behavior that this is fixing? I did some quick tests with integer keys and it seems that the regexes are getting evaluated against the integer as a string, which seems... maybe ok?

violuke:
Thanks @brianr, it's actually the ```in_array()``` bit which is the problem. The key of ```0``` is wrongly considered as in the array of ```$this->scrub_fields``` (presumably php converts the strings in the scrub_fields to ints to compare them to the keys, rather than the other way around.

See simple test of:

```
$array = [
	'something',
    'something else',
    'another',
    'this' => 'this2',
    'that' => 'that2'
];

$array2 = [
	'this',
    'that'
];

foreach ($array as $key => $value){
	if (in_array($key, $array2)){
		echo $key . ' is in $array2' . "\n";
	} else {
		echo $key . ' is not in $array2' . "\n";
	}
}
```
Which gives:
```
0 is in $array2
1 is not in $array2
2 is not in $array2
this is in $array2
that is in $array2
```
However the key of ```0``` is not in the array. I hope that makes sense?

brianr:
Thanks @violuke for your work here! I've merged #65 which seems to be a more robust fix for this issue. Will get a new release out with that shortly.

PR #65, opened 2015-07-25
Use strict comparisons when matching scrub keys

ghostal:
Better solution for bug highlighted in PR #64.

Because of the loose comparison used by default by `in_array()`, the code below causes the output that follows:
```
$fruits = ['apple', 'banana', 'cherry'];
var_dump(in_array(0, $fruits));
var_dump(in_array(0, $fruits, true));
```
```
bool(true)
bool(false)
```
In the rollbar library, this means that any zero indexed array always has its first element scrubbed.

### DIFF
--- a/src/rollbar.php
+++ b/src/rollbar.php
@@ -508,7 +508,7 @@ class RollbarNotifier {
     }
 
     protected function _key_should_be_scrubbed($key, $potential_regex_filters) {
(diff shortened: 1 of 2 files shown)


brianr:
This is great, thanks @ghostal!

I've verified that the new test fails on the old code, and passes after the fix.

Will get a new release out shortly.

PR #67, opened 2015-08-14
Add CodeIgniter instructions

Crisfole:
[ci-skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,30 @@ Rollbar::init($config, $set_exception_handler, $set_error_handler);
 
 ### For CodeIgniter Users
 
-If you are using CodeIgniter and wish to report on E_NOTICE level errors you'll have to be sure to run `Rollbar::init` from inside the Controller constructor.
+If you are using CodeIgniter you can place `Rollbar::init` in either of the two places:
+* inside the Controller's constructor
+```php
+	public function __construct()
+	{
+		Rollbar::init(array(
+			'access_token' => config_item('rollbar_access_token'),
+		 	'environment' => ENVIRONMENT
+		));
+		parent::__construct();
+	}
+```
+* `pre_system` hook
+```php
+$hook['pre_system'] = function () {
+    Rollbar::init([
+        'access_token' => config_item('rollbar_access_token'),
+        'environment' => ENVIRONMENT,
+        'root' => APPPATH . '../'
+    ]);
+};
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,6 +87,10 @@ Rollbar::init($config, $set_exception_handler, $set_error_handler);
 ?>
 ```
 
+### For CodeIgniter Users
+
+If you are using CodeIgniter and wish to report on E_NOTICE level errors you'll have to be sure to run `Rollbar::init` from inside the Controller constructor.
```
brianr: Why is this? Does this mean that `Rollbar::init` will be called on every request? That seems like it's probably papering over something else that's not working as designed...

ches:
I indeed had to initialize Rollbar in a `'post_controller_constructor'` hook to get any notifications sending to the service. My first instinct was to try something like this (`application/config/hooks.php`):

```php
$hook['pre_system'] = function () {
    Rollbar::init([
        'access_token' => config_item('rollbar_access_token'),
        'environment' => ENVIRONMENT,
        'root' => APPPATH . '../'
    ]);
};

$hook['post_system'] = function () {
    Rollbar::flush();
};
```

But it didn't work, calls to `Rollbar::report_message` or reporting exceptions silently had no effect. Changing to a `'post_controller_constructor'` hook got it working, but not with batching—I tried calling `flush()` in either `'post_system'` or `'post_controller'` to no avail. So ultimately I'm using:

```php
$hook['post_controller_constructor'] = function () {
    Rollbar::init([
        'access_token' => config_item('rollbar_access_token'),
        'environment' => ENVIRONMENT,
        'root' => APPPATH . '../',
        'batched' => false // FIXME: no luck with Rollbar::flush() in a post hook
    ]);
};
```

If anyone has advice I'm all ears, I'm a CodeIgniter newb and this is a maintenance/port project.

README.md
Crisfole: I cannot remember for the life of me. Only just now noticed this.

I think CodeIgniter was swallowing the E_NOTICE errors before Rollbar got a chance to see them. Since Rollbar behaves nicely (not swallowing) errors, sticking the init in the constructor meant that the Rollbar error handler got queued up *before* the CodeIgniter one.

I thought that in PHP all code was executed on every request by default anyway. (I'm pretty sure CodeIgniter is not a long-running app like Laravel). So yes. While it does get called on every request, it's not a terribly big deal.

Crisfole:
@ches have you tried some logging statements inside your `hook`s to verify that `init` is actually getting called?

coryvirok:
@ArturMoczulski is this still valid?

PR #69, opened 2015-09-22
Fix E_PARSE reporting level

Crisfole:
Add E_PARSE at the correct reporting level. It's a fatal level, so critical is appropriate.


coryvirok:
:+1: LGTM

brianr:
:+1:

PR #73, opened 2016-02-07
Changed all references of Exception to be \Throwable…

dlgoodchild:
so that all type of Exceptions in PHP7 can be caught by Rollbar.
Maybe you want to start a php7 repo so there's no clash?

### DIFF
--- a/src/rollbar.php
+++ b/src/rollbar.php
@@ -4,6 +4,10 @@
  *
  * Unless you need multiple RollbarNotifier instances in the same project, use this.
  */
+if ( !defined( 'BASE_EXCEPTION' ) ) {
+	define( 'BASE_EXCEPTION', version_compare( phpversion(), '7.0', '<' )? '\Exception': '\Throwable' );
+}
+
 class Rollbar {
     /** @var RollbarNotifier */
     public static $instance = null;
@@ -177,8 +181,8 @@ class RollbarNotifier {
 
     public function report_exception($exc, $extra_data = null, $payload_data = null) {
         try {
-            if (!$exc instanceof \Throwable) {
-                throw new Exception('Report exception requires an instance of \Throwable.');
+            if ( !is_a( $exc, BASE_EXCEPTION ) ) {
+                throw new Exception(sprintf('Report exception requires an instance of %s.', BASE_EXCEPTION ));
             }
 
             return $this->_report_exception($exc, $extra_data, $payload_data);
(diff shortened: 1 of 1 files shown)


dlgoodchild:
http://php.net/manual/en/language.errors.php7.php

mbardelmeijer:
:+1: just ran into this issue.

ikovalyov:
:+1:

ikovalyov:
Do you think we can check php version and call different types of exceptions depending on it? Then tests won't fail.

Crisfole:
Hey all, thanks for the feedback, sorry for the sluggish response here.

What do most other PHP libraries do to handle varying language versions? Is it typically a fork, custom behavior based on PHP version, or does Composer have 'runtime version' behaviors we can take advantage? We want to be as idiomatic as possible here...

dlgoodchild:
My implementation was done fast to get working again. You could tag with a leap version so people know to know the v1 is for php7 use? Or, as suggested, yeah we could check the version and if < 7, throw Exception, otherwise throw Throwable.

nocive:
I'd go for the latter as it's simply easier to integrate, maintain and, to my knowledge, has no downsides.
Just my two cents :)

Crisfole:
For the time being is it possible to use an alias with an import here? Something like (PSEUDOCODE AHEAD):

```php
if (isPhp7OrHigher()) {
    use \Throwable as Exception;
}
```

If it worked as expected it would do the same thing as this PR without changing the behavior in previous versions.

nocive:
@Crisfole that's not possible, PHP's use statements cannot be block scoped because the importing is done at compile time and not runtime.

dlgoodchild:
Yeah @nocive is right, that wouldn't be possible. I've just updated my fork which has a version compare and defines the base exception class accordingly. Further, replaced instances of "instanceof" operator with is_a() function as it can accept a string/variable as a comparison argument unlike instanceof. Build passes.

Crisfole:
I see that, LGTM! Do you mind just applying the second commit to this branch so we can keep the discussion history attached to the code that gets merged?

@brianr Can you take a peek at https://github.com/rollbar/rollbar-php/pull/83?

dlgoodchild:
@Crisfole done

Crisfole:
@dlgoodchild As soon as @brianr gets a chance to review too we can merge. I'd like to up the version number with a few of these PRs merged in.

brianr:
Looks good to me, thanks all. We'll need to do a version bump; I'll coordinate with @Crisfole on that.

Crisfole:
Great! I'm going to merge this along with a few other PRs later today and then I'll bump the version.

dlgoodchild:
Awesome! Nice one, now I don't need my fork :D

Crisfole:
Lemme know if I messed anything up...I've never published with composer before.

dlgoodchild:
Sorry, but something's not quite right. I'm looking at the changes involved in 0.16.0 and I'm not seeing this commit; which explains why when I pull 0.16.0 via composer I'm not seeing the changes.

nocive:
Here's why: https://github.com/rollbar/rollbar-php/compare/v0.16.0...master
Tag was not updated after the merge

PR #74, opened 2016-03-11
Add constants to the class and replace static text with constants

bocharsky-bw:


src/rollbar.php
```diff
@@ -5,6 +5,11 @@
  * Unless you need multiple RollbarNotifier instances in the same project, use this.
  */
 class Rollbar {
+    const LEVEL_ERROR = 'error';
+    const LEVEL_CRITICAL = 'critical';
+    const LEVEL_WARNING = 'warning';
+    const LEVEL_INFO = 'info';
+
```
Crisfole: My first thought about this was "It ought to be an enum", but PHP doesn't have those.

Could you instead use an abstract class with those const values instead? (per: http://stackoverflow.com/a/254543/456188).

Something like:

```php
abstract class Level {
  const LEVEL_ERROR = 'error';
  const LEVEL_CRITICAL = 'critical';
  const LEVEL_WARNING = 'warning';
  const LEVEL_INFO = 'info';
}
```

src/rollbar.php
```diff
@@ -5,6 +5,11 @@
  * Unless you need multiple RollbarNotifier instances in the same project, use this.
  */
 class Rollbar {
+    const LEVEL_ERROR = 'error';
+    const LEVEL_CRITICAL = 'critical';
+    const LEVEL_WARNING = 'warning';
```
Crisfole: Missing 'debug'

Crisfole: I recognize it's not in the original, but this is a great opportunity to fix that!

Crisfole:
If you can fix these and get the tests passing I'd love to get this merged!

PS: you can run the unit tests locally with `phpunit`

bocharsky-bw:
@Crisfole your comments make sense! I've updated this PR

Crisfole:
Looks fantastic! Waiting on the tests, if/when they pass I'd say this is a go.

@brianr eyes?

brianr:
LGTM

Crisfole:
Merging later today along with a few others and a minor version bump.

PR #75, opened 2016-03-16
Fix the E_PARSE error level, forgot a `break`

Crisfole:
Add missing break statement, left E_PARSE errors in `info` level.


brianr:
LGTM

Issue #45, opened 2015-02-16
checkIgnore for PHP


johanna:
It would be great if the config option checkIgnore was implemented for PHP!

jshah4517:
+1

dlsteuer:
added in #82

PR #82, opened 2016-04-12
Added checkIgnore config option. (#45)

bytestream:
Copied over the [checkIgnore](https://rollbar.com/docs/filter-or-ignore-errors-from-bots/) functionality from https://github.com/rollbar/rollbar.js/blob/a072cb78a9056ebded63879e6eec473d6ff1752d/release/rollbar-1.8.5.js

### DIFF
--- a/src/rollbar.php
+++ b/src/rollbar.php
@@ -249,7 +249,7 @@ class RollbarNotifier {
     {
         try {
             if (is_callable($this->checkIgnore)
-                && call_user_func_array($this->checkIgnore, [$isUncaught,$caller_args,$payload])
+                && call_user_func_array($this->checkIgnore, array($isUncaught,$caller_args,$payload))
             ) {
                 $this->log_info('This item was not sent to Rollbar because it was ignored. '
                     . 'This can happen if a custom checkIgnore() function was used.');
--- a/src/rollbar.php
+++ b/src/rollbar.php
@@ -78,6 +78,36 @@ class Rollbar {
     }
 }
 
+class RollbarException {
+    private $level;
+    private $message;
+    private $err;
+    private $custom;
+
+    public function __construct($level, $message, Exception $err = null, $custom = null) {
+        $this->level = $level;
+        $this->message = $message;
+        $this->err = $err;
+        $this->custom = $custom;
+    }
+
+    public function getLevel() {
+        return $this->level;
+    }
+
+    public function getMessage() {
+        return $this->message;
+    }
+
+    public function getException() {
+        return $this->err;
+    }
+
+    public function getCustom() {
+        return $this->custom;
+    }
+}
+
 // Send errors that have these levels
 if (!defined('ROLLBAR_INCLUDED_ERRNO_BITMASK')) {
     define('ROLLBAR_INCLUDED_ERRNO_BITMASK', E_ERROR | E_WARNING | E_PARSE | E_CORE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR);
@@ -240,12 +270,12 @@ class RollbarNotifier {
     /**
      * Run the checkIgnore function and determine whether to send the Exception to the API or not.
      *
-     * @param  bool  $isUncaught
-     * @param  mixed $caller_args
-     * @param  mixed $payload
+     * @param  bool             $isUncaught
+     * @param  RollbarException $caller_args [level, message, err, custom]
+     * @param  array            $payload     Data being sent to the API
      * @return bool
      */
-    protected function _shouldIgnore($isUncaught, $caller_args, $payload)
+    protected function _shouldIgnore($isUncaught, RollbarException $caller_args, array $payload)
     {
         try {
             if (is_callable($this->checkIgnore)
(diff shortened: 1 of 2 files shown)


nocive:
:+1:

src/rollbar.php
```diff
@@ -237,6 +238,35 @@ public function queueSize() {
...
+     * @param  bool  $isUncaught
+     * @param  mixed $caller_args
+     * @param  mixed $payload
+     * @return bool
+     */
+    protected function _shouldIgnore($isUncaught, $caller_args, $payload)
+    {
+        try {
+            if (is_callable($this->checkIgnore)
+                && call_user_func_array($this->checkIgnore, [$isUncaught,$caller_args,$payload])
```
Crisfole: http://php.net/manual/en/function.func-get-args.php

I know that this is a pretty 'boring' function that's hopefully rarely changed, but could you use `func_get_args` here (and if not, why not?). It would make sure that no matter what we changed about the function arguments it would still work.

src/rollbar.php
```diff
@@ -237,6 +268,35 @@ public function queueSize() {
...
+     * @param  bool             $isUncaught
+     * @param  RollbarException $caller_args [level, message, err, custom]
+     * @param  array            $payload     Data being sent to the API
+     * @return bool
+     */
+    protected function _shouldIgnore($isUncaught, RollbarException $caller_args, array $payload)
+    {
+        try {
+            if (is_callable($this->checkIgnore)
+                && call_user_func_array($this->checkIgnore, array($isUncaught,$caller_args,$payload))
```
Crisfole: Hm. So `RollbarException` seems to be duplicating a lot of stuff that's already present in `$payload`. Is there a notable difference?

Crisfole:
On a more general level, would this function allows users to tweak the payload before it gets sent?

It's not necessarily a bad thing, but mutating can have annoying side effects. Is there an easy, fast deep clone mechanism in PHP? That could make this 'pure' (apart from its arguments) by passing a cloned version of the `$payload` to the checkIgnore function.

Note: this is why there's a separate `transform` config function in Rollbar.js.

src/rollbar.php
bytestream: I literally just copied the functionality and tried to ensure it was consistent with the JS version. I agree it doesn't really make sense; I wasn't sure how to handle the two cases where there is not an exception:

* Logging a message
* Logging a PHP error

bytestream:
At the moment they can't tweak the payload no.

Crisfole:
@bytestream what prevents them from tweaking it? Couldn't I just say `$payload["data"] = "LOOKIE HERE!!!"` in my callback? (Here I may be revealing my PHP ignorance).

bytestream:
@Crisfole PHP passes function args by value not by reference :) http://php.net/manual/en/functions.arguments.php

Crisfole:
Right, but the 'value' of most object types (arrays and maps included) in most languages is simply the reference to the object. Does PHP not work this way?

nocive:
@Crisfole only for objects, otherwise you have to explicitly say you want to use them by reference with ```&```

Crisfole:
:+1: I'd say this is close to good to go once we make sure that the `Exception`, Error data, and message are availabe in the second argument in some way shape or form. (From IRC channel)

bytestream:
@Crisfole done ^

src/rollbar.php
```diff
@@ -476,7 +464,8 @@ protected function _report_php_error($errno, $errstr, $errfile, $errline) {
         $payload = $this->build_payload($data);
 
         // Determine whether to send the request to the API.
-        if ($this->_shouldIgnore(true, new RollbarException($level, $errstr), $payload)) {
+        $exception = new ErrorException($error_class, 0, $errno, $errfile, $errline);
```
Crisfole: :+1: That's great! I didn't know `ErrorException` existed. That's a nice solution.

Crisfole:
K, last thought is that we probably don't need the `new RollbarException($e->getMessage(), $e)` (the `getMessage` portion in particular.

If people want to call `getMessage` they can be our guest but they may not need it, and it makes it just the tiniest bit harder to tell if a `RollbarException` is for a `report_message` or `report_exception`.

It *might* be nice to have a third parameter for `ErrorException` instead of using a single exception parameter. It might change how you want to handle the ignoring.

Thoguhts?

bytestream:
You can differentiate between a message and an exception by just checking if ````getException()```` returns null. Changing it to how you suggest is really just the same thing, you still have to check which params are set. This is how you would do it at the moment:

````php
public function ignore($isUncaught, RollbarException $exception, $payload)
{
	if (is_null($exception)) {
		// we have a message
	} else {
		if ($exception instanceof ErrorException) {
			// we have a php error
		} else {
			// we have a normal error
		}
	}
}
````

The OO way is more long winded, arguably a nicer, but would be this (untested):

````php
interface RollbarException {
	public function getMessage();
}

class RollbarPhpError extends ErrorException implements RollbarException
{
	//
}

class RollbarMessage implements RollbarException
{
	public function getMessage()
	{
		//
	}
}

class RollbarUncaughtException extends Exception implements RollbarException
{
	//
}

public function ignore($isUncaught, RollbarException $exception, $payload)
{
	if ($exception instanceof RollbarPhpError) {
		// 
	} elseif ($exception instanceof RollbarMessage) {
		//
	} elseif ($exception instanceof RollbarUncaughtException) {
		//
	} else {
		// getMessage()
	}
}
````

Crisfole:
I think I'm good with it the way it is, then. @brianr what do you think?

@bytestream Once we get a second pair of eyes on this we can merge and up the version number.

dlsteuer:
Looks good to me, merging in.

PR #89, opened 2016-05-13
PSR-0 support

vit1251:
Hello,

     I use modern PHP and have issue with classmap loader in composer.
     I make some small changes to support PSR-0 support.
     Please review this changes.

Thanx.


composer.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "rollbar/rollbar",
+  "name": "make-test/rollbar",
```
Crisfole: Can you undo this change?

composer.json
```diff
@@ -1,10 +1,10 @@
...
   "description": "Monitors errors and exceptions and reports them to Rollbar",
   "type": "library",
   "keywords": ["logging", "debugging", "monitoring", "errors", "exceptions"],
   "license": "MIT",
-  "homepage": "http://github.com/rollbar/rollbar-php",
+  "homepage": "http://github.com/make-test/rollbar-php",
```
Crisfole: And this one

Crisfole:
Vitold: Thanks for this. We're actually mid-refactor for a 1.0 release which is PSR 1-4 compliant. The early release version should be ready either today or early next week. Will that work for your purposes?

PR #91, opened 2016-06-03
Monolog docs

dlsteuer:


README.md
```diff
@@ -5,6 +5,27 @@
 This library detects errors and exceptions in your application and reports them to [Rollbar](https://rollbar.com) for alerts, reporting, and analysis.
 
 <!-- Sub:[TOC] -->
```
brianr: Since you've added a TOC directly into the readme, we should remove the above line - it gets expanded on rollbar.com/docs to build its own TOC.

brianr:
LGTM

PR #92, opened 2016-06-04
making flush() check for instance first

asuth:


dlsteuer:
Thanks @asuth looks good to me.

PR #94, opened 2016-06-14
Refactor1.0

dlsteuer:
@brianr @coryvirok this is quite a big change, so will need some careful review


src/Config.php
```diff
@@ -0,0 +1,343 @@
...
+
+        if (isset($c['included_errno'])) {
+            $this->included_errno = $c['included_errno'];
+        }
+    }
+
+    private function setAccessToken($c)
+    {
+        if (isset($_ENV['ROLLBAR_ACCESS_TOKEN']) && !isset($config['access_token'])) {
+            $config['access_token'] = $_ENV['ROLLBAR_ACCESS_TOKEN'];
```
chrisbarmonde: Looks like a little bit of variable confusion here. Param in the function is `$c`, but here and in the conditional we're checking for `$config`.

src/Config.php
```diff
@@ -0,0 +1,343 @@
...
+            $this->error_sample_rates = $configArray['error_sample_rates'];
+        }
+
+        $levels = array(E_WARNING, E_NOTICE, E_USER_ERROR, E_USER_WARNING,
+            E_USER_NOTICE, E_STRICT, E_RECOVERABLE_ERROR);
+        // PHP 5.3.0
+        if (defined('E_DEPRECATED')) {
+            $levels = array_merge($levels, array(E_DEPRECATED, E_USER_DEPRECATED));
+        }
+        $curr = 1;
```
chrisbarmonde: Don't think this var is used

src/Config.php
```diff
@@ -0,0 +1,343 @@
...
+
+    private function setTransformer($c)
+    {
+        $expected = "Rollbar\TransformerInterface";
+        $this->setupWithOptions($c, "transformer", $expected);
+    }
+
+    private function setMinimumLevel($c)
+    {
+        if (!isset($c['minimumLevel']) || is_null($c['minimumLevel'])) {
```
chrisbarmonde: This could just be `if (empty($c['minimumLevel'])) {`

src/DataBuilder.php
```diff
@@ -0,0 +1,803 @@
...
+        }
+    }
+
+    protected function getOrCall($name, $level, $toLog, $context)
+    {
+        if (is_callable($this->$name)) {
+            try {
+                return $this->$name($level, $toLog, $context);
+            } catch (\Exception $e) {
+                // TODO Report the configuration error.
```
chrisbarmonde: Something we still want to do?

src/DataBuilder.php
```diff
@@ -0,0 +1,803 @@
...
+    }
+
+    public function makeFrames($exception)
+    {
+        $frames = array();
+        foreach ($this->getTrace($exception) as $frameInfo) {
+            $filename = Utilities::coalesce($this->tryGet($frameInfo, 'file'), '<internal>');
+            $lineno = Utilities::coalesce($this->tryGet($frameInfo, 'line'), 0);
+            $method = $frameInfo['function'];
+            // TODO 4 (arguments are in $frame)
```
chrisbarmonde: ??

src/Utilities.php
```diff
@@ -0,0 +1,120 @@
...
+        foreach ($values as $key => $val) {
+            if ($val) {
+                return $val;
+            }
+        }
+        return null;
+    }
+
+    // Modified from: http://stackoverflow.com/a/1176023/456188
+    public static function pascaleToCamel($input)
```
chrisbarmonde: The biggest of nitpicks, but it'd really need to be called `pascalToCamel`.

tests/BackwardsCompatibilityConfigTest.php
```diff
@@ -0,0 +1,64 @@
...
+            'timeout' => 10,
+            'report_suppressed' => true,
+            'use_error_reporting' => true,
+            'proxy' => array(
+                'address' => '127.0.0.1:8080',
+                'username' => 'my_user',
+                'password' => 'my_password'
+            )
+        ));
+    }
```
chrisbarmonde: Should we actually be validating the value of any of the backwards compat fields in this test? Or do they like throw exceptions if they're not set properly? At the very least, it might be worth highlighting *which* of the fields are backwards compatible

chrisbarmonde:
Well that's about as useful as I'm gonna be. I will say, in my time with PHP, I found it was almost always better to just use `empty` as opposed to `isset` or `is_null`. I feel like in 99.999% of cases, you just care whether the object exists or not (which includes it not being `null` already). They have a big ol' table with what each of the different functions returns here: http://php.net/manual/en/types.comparisons.php

tests/BackwardsCompatibilityConfigTest.php
dlsteuer: I put that in the first place to see what would blow up when passed a valid config array from the old notifier. Then I left it in as a bit of checklist.

src/Config.php
dlsteuer: Thanks @chrisbarmonde fixed with a test now.

src/Config.php
dlsteuer: It's used on line 65 to set the error levels that don't exist to 1, so they always get sent.

src/DataBuilder.php
dlsteuer: At some point it would be good to add the arguments into the payload at this point we don't

dlsteuer:
with the whole empty vs isset you have to really assess what empty means at the comparison point, for example "0" is that empty or not in the particular case where you are testing.

src/Config.php
chrisbarmonde: Wowzers, I am blind as a bat. 👀  Sorry about that

PR #109, opened 2016-10-19
Check global var

truerainbow:


ArturMoczulski:
@truerainbow thanks for the pull request. Changes looks great. Do you have any specific reason why you would like to see this changes pull into master? Other than checking global vars is just a good practice?

PR #110, opened 2016-11-14
Specifying latest PSR log to avoid minimum-stability issues

manchuck:
If a 3rd party has minimum-stability: stable set in their composer.json, they are un able to install rollbar-php.  This is due to rollbar-php requiring psr/log @ dev-master which is not a stable release.

It is best practice to include a specific version for a vendor library

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
 
   "require": {
     "ext-curl": "*",
-    "psr/log": "dev-master",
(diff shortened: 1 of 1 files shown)


manchuck:
This is resolved with #305

manchuck:
Closed it by mistake

rokob:
Seems reasonable to me

PR #111, opened 2016-11-14
Fix TypeError(s) and serialize ExceptionInfo

MichaelMackus:
Fixes #100, fixes #106

PR #112, opened 2016-11-16
Fix scrubber

MichaelMackus:
Branch split of #111, specifically to fix the DataBuilder::scrub method.

### DIFF
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -756,7 +756,7 @@ class DataBuilder implements DataBuilderInterface
 
         $scrubber = function (&$val, $key) use ($fields, $replacement, $arr) {
 
-            if (key_exists($key, $arr)) {
+            if (key_exists($key, $fields)) {
                 $val = str_repeat($replacement, 8);
             }
         };
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -754,13 +754,18 @@ class DataBuilder implements DataBuilderInterface
             return null;
         }
 
-        $scrubber = function (&$val, $key) use ($fields, $replacement, $arr) {
-
+        $scrubber = function (&$val, $key) use ($fields, $replacement, &$scrubber) {
             if (in_array($key, $fields)) {
                 $val = str_repeat($replacement, 8);
             }
+
+            if (is_array($val)) {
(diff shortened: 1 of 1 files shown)


MichaelMackus:
Just to be clear, from what I could tell from initial usage the "scrubFields" param was incredibly broken. It seems like it just ended up scrubbing all the fields, no matter the "scrubFields" param.

I've gone through the scrubber function and simply fixed some parameters & other things to ensure the "scrubFields" param is recognized.

MichaelMackus:
[Here's](http://php.net/manual/en/function.in-array.php#106319) a link I found explaining some reasoning behind the previous commit.

MichaelMackus:
@rokob any input? This definitely fixes a lot of issues with the previous scrubber functionality.

src/DataBuilder.php
```diff
@@ -754,13 +754,18 @@ protected function scrub($arr, $fields, $replacement = '*')
...
+        $scrubber = function (&$val, $key) use ($fields, $replacement, &$scrubber) {
             if (in_array($key, $fields)) {
                 $val = str_repeat($replacement, 8);
             }
+
+            if (is_array($val)) {
+                array_walk($val, $scrubber);
+            }
         };
-        array_walk_recursive($arr, $scrubber);
```
ArturMoczulski: This is confusing to me. I don't understand the reasoning behind doing a manual recursion versus array_walk_recursive()

ArturMoczulski: Oh, I just saw @MichaelMackus comment above. I think that making the code to error out when passing other types than array is a good idea, since they are not going to get scrubbed anyway. My reasoning is that if I pass some data to be scrubbed (e.g. an object) and the library doesn't support this type, it's a pretty critical issue. We don't want the users to be passing objects and thinking they got scrubbed when they didn't.

ArturMoczulski: Argument I'm making here is that if a data type not supported by the scrubbing method is passed we should at least throw an exception and make it very obvious.

ArturMoczulski:
Improvements @MichaelMackus did are looking great. I'm going to take his work, use it as a base for my branch and expand on them.

PR #116, opened 2017-02-02
Fix Uncaught TypeError: Argument 3 passed to Rollbar\RollbarLogger::log() must be of the type array

nocive:


nocive:
Closing as the same problem seems to have been addressed in https://github.com/rollbar/rollbar-php/pull/111

Issue #122, opened 2017-04-03
Test scrubbing


rokob:
As per https://github.com/rollbar/rollbar-php/pull/112 scrubbing does not work correctly. There is one test for scrubbing and it obviously does not cover a reasonable amount of surface area. We need better tests around how things are scrubbed.

rokob:
https://github.com/rollbar/rollbar-php/blob/master/tests/DataBuilderTest.php#L152-L163 is the one place that scrubbing is currently tested.

rokob:
I would check out the comments on https://github.com/rollbar/rollbar-php/pull/112 as to the reasoning for those proposed changes to scrubbing are and what the expected behavior is. In general, if a have a key in scrub_fields, we want to replace the value with `********` but we want to do that within recursive structures as well.

ArturMoczulski:
This is covered by PR https://github.com/rollbar/rollbar-php/pull/126

Issue #52, opened 2015-04-30
Option to exclude cookie data in request


dataground:
The scrub_fields param should also look at $_COOKIE to exclude custom session cookies containing sensitive information. Maybe a scrub_cookies param?

dataground:
Party related to https://github.com/rollbar/rollbar-php/issues/15

ArturMoczulski:
This is covered by PR #126

ArturMoczulski:
In master we don't include any $_COOKIE data in the request. This is not relevant anymore.

Issue #31, opened 2014-06-24
Do scrub fields apply to $extra_data


nodesocket:
When passing custom `$extra_data` into `report_message()` do the settings scrub fields apply, i.e. mask those values?

ArturMoczulski:
This is covered by PR #126

Issue #15, opened 2014-03-04
Remove session id from posted cookie


drjayvee:
In 5.x, we extended `RollbarNotifier::build_request_data` to strip the session id from the cookie data. In 6.x, that method and the member `_request_data` have been made `private`, so we cannot touch that. (Which is exactly why I dislike "private by default")

`RollbarNotifier::scrub_fields` only applies to `$_POST` and `$_SESSION`, so that can't be used either.

I would propose that Rollbar strip the sesion id (`$_COOKIE[session_name()]`) to prevent session ids from leaking, which is a security risk (because it enables session hijacking). More generally, if `scrub_fields` _were_ applied to headers, users could filter out whatever they want to.

dataground:
Related to https://github.com/rollbar/rollbar-php/issues/52

ArturMoczulski:
In master we don't include any $_COOKIE information in the payload. This issue is not relevant anymore.

PR #126, opened 2017-04-11
GitHub Issue #122: fix scrubbing and provide test coverage

ArturMoczulski:
This pull request covers the following GitHub issues: #122 #52 #31 #15 
It's still work in progress but I encourage collaboration.

It includes work on the following:

- provide test coverage for scrubbing of $_GET, $_POST, query string, request extras, headers, request body context, $_SESSION, server extras, Data::getCustom(), args and kwargs in frames as mentioned in POST_FORMAT.md, ::log() $extra param
- add recursive scrubbing,
- add recursive scrubbing of query strings,
- remove code duplication and use a single point of scrubbing logic,
- refactor placement of scrubbing logic to follow the Python SDK standard


ArturMoczulski:
This is ready for review

src/DataBuilder.php
```diff
@@ -743,38 +727,94 @@ protected function getBaseException()
...
+                    $data
+                );
+            }
+        }
+        
+        return $data;
+    }
 
-    protected function scrub($arr, $fields, $replacement = '*')
+    public function scrubArray(&$arr, $replacement = '*')
```
coryvirok: Do these really need to be public?

ArturMoczulski:
@coryvirok Good point. I made them public earlier for unit testing, but since tests for scrub() already cover the same data structures no need to expose those methods. Just pushed a commit with a change fo this.

coryvirok:
Is there anything blocking this PR from being merged? cc @rokob

src/DataBuilder.php
```diff
@@ -743,38 +727,94 @@ protected function getBaseException()
...
 
-        $scrubber = function (&$val, $key) use ($fields, $replacement, $arr) {
-
-            if (key_exists($key, $arr)) {
+        $scrubber = function (&$val, $key) use ($fields, $replacement, &$scrubber, $dataBuilder) {
+            if (in_array($key, $fields, true)) {
                 $val = str_repeat($replacement, 8);
             }
+
+            $val = $dataBuilder->scrub($val, $replacement);
```
rokob: Why not have this in an else block? It would seem that it is okay to not have it in an else because $val just ends up as the string '********' which scrub will just ignore, but why not avoid that call to scrub altogether?

tests/DataBuilderTest.php
```diff
@@ -28,6 +24,9 @@ public function setUp()
     public function testMakeData()
     {
         $output = $this->dataBuilder->makeData(Level::fromName('error'), "testing", array());
+        /**
+         * @todo test for scrub data in makeData()
```
rokob: Do we need this?

rokob:
LGTM besides some minor comments

src/DataBuilder.php
ArturMoczulski: Good point

tests/DataBuilderTest.php
ArturMoczulski: No. I'm going to clean this up.

ArturMoczulski:
@rokob @coryvirok 
code climate brought up this code:
https://github.com/ArturMoczulski/rollbar-php/blob/github-122-scrubbing/src/RollbarLogger.php#L35-L40

with this message:
_The method log uses an else expression. Else is never necessary and you can simplify the code to work without else._

Do you guys have any suggestion to how to structure this code to not use the else statement? The only way without an else block I can think of is, but that's worse in my opinion:
`
$response = new Response(0, "Ignored");        
if (!$this->config->checkIgnored($payload, $accessToken, $toLog)) {
            $scrubbed = $this->scrub($payload);
            $response = $this->config->send($scrubbed, $accessToken);
}
`

coryvirok:
I would disregard the CodeClimate suggestion in this instance. As for the other items from CC, they're not high priority as long as we have good coverage. i.e. we can do a cleanup round after release.

rokob:
Also, "never use else" seems like terrible advice.

ArturMoczulski:
All suggestions addressed so if there is nothing more this can be merged into master.

rokob:
👍

Issue #70, opened 2015-10-29
[Feature Request]: Try-catch block around 'person_fn'


davidsteinsland:
Would be nice if the call to `person_fn` callable would be wrapped in a try-catch block, and swallow the exception (logging it with `log_error`).

The problem I had:
- PHP error was generated from PHP-PDO layer (connection warning)
- `person_fn` was called, which checks if a session exists and returns user information
- Sessions are stored in database, meaning that the session checking would again result in a PDO warning
- This would fail, as `report_php_error` would fail with "Exception while reporting php error".

ArturMoczulski:
This is ready for review in https://github.com/rollbar/rollbar-php/pull/134

rokob:
Merged

PR #134, opened 2017-04-24
GitHub Issue #70: person fn

ArturMoczulski:
Catch, log and swallow exceptions coming from `person_fn`


src/DataBuilder.php
```diff
@@ -609,7 +611,12 @@ protected function getPerson()
     {
         $personData = $this->person;
         if (!isset($personData) && is_callable($this->personFunc)) {
-            $personData = call_user_func($this->personFunc);
+            try {
+                $personData = call_user_func($this->personFunc);
+            } catch (PersonFuncException $exception) {
```
rokob: Doesn't this only handle the specific case of someone throwing this particular type of exception? That doesn't seem to handle the general case of errors happening in user code. It seems like we should swallow all exceptions regardless of type coming from user code.

ArturMoczulski:
@rokob Obviously. Thanks for pointing this out. Fixed to swallow all.

rokob:
We might want to address that codeclimate issue, but I'm good with this change, so merging.

Issue #132, opened 2017-04-17
Add backwards compatibility with 0.18.*


ArturMoczulski:
Add ::report_message and ::report_exception to Rollbar class for backwards compatibility with 0.18.*. These should be wrappers around the new ::log solution and marked as deprecated.

ArturMoczulski:
Rollbar::flush() also seems to be deprecated. From what I see the batching functionality is just not there in 1.0.0. anymore. I can think of two approaches to go about this:

1. Leave Rollbar::flush() for backwards compatibility and have it do nothing (so users who have those function calls in their code currently do not start getting fatal errors - missing method).
2. Remove Rollbar::flush() completely which will force fatal errors in our user's code and force them to remove these calls if they have any.

@coryvirok @rokob what do you think?

brianr:
Strong vote for option 1. Upgrading to 1.0 shouldn't break customers' code.

rokob:
Merged

PR #136, opened 2017-04-27
GitHub Issue #132: backwards compatibility with 0.18.2

ArturMoczulski:
Added the following methods to \Rollbar\Rollbar:

- report_message() - wrapper for log()
- report_exception() - wrapper for log()
- report_fatal_error() - wrapper for fatalHandler()
- report_php_error() - wrapper for errorHandler()
- flush() - empty method to not cause fatal errors in users' codebases

Added tests in tests/RollbarTest.php to check backward compatibility.

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -101,4 +101,74 @@ class Rollbar
     {
         return new Response(0, "Rollbar Not Initialized");
     }
+    
+    /**
+     * Below methods are deprecated and still available only for backwards
+     * compatibility. If you're still using them in your application, please
+     * transition to using the ::log method as soon as possible.
+     */
+    
+    /**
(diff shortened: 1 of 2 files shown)


ArturMoczulski:
All of the code climate issues I see are not really avoidable for this backward compatibility so I'm going to ignore those.

ArturMoczulski:
This is ready for review

rokob:
This looks good to me. Is this: "Avoid using static access to class '\Rollbar\Payload\Level'" something we should care about?

rokob:
I looked into the static access warning, I think there might be something we could do to refactor the way levels are defined and used, but it is not important for what we are doing at the moment. So I am going to merge this.

Issue #43, opened 2015-01-16
Return result from API call


digilist:
I'd like to know whether my log request to rollbar was successful or not, so that I could log to a file or send a mail (or react in other ways), when the request failed.
Currently, the result is only logged via your own logger (which isn't even PSR-3 compliant).

nodesocket:
@digilist perhaps you could take a crack at submitting a PR then?

digilist:
Yeah I can do this, I thought I'd look for feedback and other thoughts first. Maybe there are reasons not to do this.

brianr:
@digilist we'd be very open to this. If it's going to be a lot of code, please share out the design first for comment; if it's not a lot then just going straight to a PR is great.

ArturMoczulski:
This is already available in v1.0.0 through `$response->wasSuccessful()`

I added an example of how to use this functionality to README.md in the PR

PR #141, opened 2017-04-29
Github Issue #43: Return result from API call

ArturMoczulski:
Functionality request in #43 is already available in v1.0.0

This pull request is just an amendment to README.md with an example of how to use it


rokob:
👍

Issue #72, opened 2015-12-11
status 200 when using set_exception_handler


oadam:
I'm using rollbar with great success but today I've noticed something that bothers me.

My code is using exception to report invalid GET parameters.
In dev, when I call an invalid URL, I get a 500 error and the error is logged by php.
In prod, where rollbar is enabled, I obtain a blank page with a 200 status.

Having a 200 status can be an issue for a number of reasons (for example because it means a search engine could index pages whose URL is invalid).

In my opinion, most people would like the server to return a 5XX status, so it makes sense to implement this in rollbar's `error_handler`.

Do you think that's a good idea ?

Crisfole:
Hey there, sorry for the long delay!

I don't think we're purposely setting the status to 200. On my end I still get a 500 when I report something to Rollbar.

Are you sure this is the application and not your server in production? (Apache can easily do stuff like this, for example).

nocive:
Seeing this exact same behaviour too after switching to php7.
It's definitely a rollbar related problem because when rollbar is not enabled everything works as expected.

nocive:
See also https://github.com/getsentry/sentry-php/issues/280 for what seems like the same issue.

ArturMoczulski:
This is ready for review in PR #143

rokob:
Fixed in #143

PR #143, opened 2017-04-29
GitHub Issue #72: status 200 when using set_exception_handler

ArturMoczulski:
This changes the exception handler to trigger logging with Rollbar and follow with the default PHP global exception handler instead of overriding it. This effectively brings back the default PHP behavior for uncaught exceptions which is:

```
if (php_ini.display_errors == 1) {
    show error info
    HTTP RESPONSE CODE 200
} else {
    HTTP RESPONSE CODE 500
}
```

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -53,7 +53,15 @@ class Rollbar
     
     public static function exceptionHandler($exception)
     {
-        return self::log(Level::error(), $exception);
+        self::log(Level::error(), $exception);
+        
+        if (ini_get('display_errors')) {
+            restore_exception_handler();
+            throw $exception;
+        } else {
+            header($_SERVER['SERVER_PROTOCOL'] . ' 500 Internal Server Error', true, 500);
+            exit(1);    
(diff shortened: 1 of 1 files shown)


rokob:
This seems like a straight forward change, will this break any existing code? It seems like we went from catching and handling errors to now throwing an exception.

ArturMoczulski:
@rokob That's a good question. Theoretically, it might "break" some existing code. By "break" I mean we are not going to hide exceptions anymore. The thing is, those would be exceptions that were there already. Now we change a blank page with a 200 response code into a 500 response code or a standard verbose PHP error. I would say this is not "breaking" existing code, because a blank page with a 200 response code is just as broken as a 500 response code imho. Unless I'm missing something here of course.

ArturMoczulski:
What this does is bringing back the default PHP behavior + Rollbar logging as opposed to mask errors as blank 200 rc's.

rokob:
Okay right that is what I thought was happening. Merging this then.

Issue #53, opened 2015-05-14
Option to capture stack trace in report_message()


brianr:
It would be nice if calls to Rollbar::report_message() could also capture the stack trace.

This should probably be available both as a per-call param, and as a config param.

ArturMoczulski:
This is ready for review in PR #145

rokob:
Merged

PR #145, opened 2017-04-29
GitHub Issue #53: Option to capture stack trace in report_message()

ArturMoczulski:
Added backtrace to message logged with `Rollbar::log(string)`. This is going to be saved to the Rollbar API as `data.body.message.backtrace`

Ideally, it would be great to also update Rollbar UI to support displaying those stack traces in a human-friendly format. Currently, it's just a stack trace dump as string from PHP's `debug_backtrace()`

I experimented a bit with sending the message wrapped in an exception which would display the trace info in a human-friendly way by the default. However, doing that removes the ability to inspect `data.body.message.extra`

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,6 +418,12 @@ $config['proxy'] = array(
 Default: No proxy
 </dd>
 
+<dt>send_message_trace</dt>
+<dd>Should backtrace be include with string messages reported to Rollbar
+
+Default: `false`
+</dd>
+
 </dl>
 
 Example use of error_sample_rates:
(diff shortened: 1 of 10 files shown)


rokob:
I don't think we want this by default, it should be enabled via a configuration option.

ArturMoczulski:
Added a configuration option to enable / disable this `send_message_trace` which is set to `false` by default.

There is no way to pass this per-call. It's much cleaner to use the config object for this.

rokob:
LGTM, I agree that the configuration option is the way to go rather than adding a per-call option as well.

PR #152, opened 2017-05-04
Fix check ignore [Fixes #151]

### LINKED ISSUE #151: checkIgnore doesn't pass parameters
https://github.com/rollbar/rollbar-php/blob/master/src/Config.php#L292

I'm trying to implement checkIgnore on 1.0.1 and i am getting warnings about missing paramaters

My config entry
```
checkIgnore' => function ($isUncaught, $exception, $payload) {
                print_r($exception);
                return false;
}
```
I don't see where this is passing any parameters? The documentation says it should be passing 3 parameters
       ```
 if (isset($this->checkIgnore) && call_user_func($this->checkIgnore)) {
            return true;
        }
```

rokob:
According to our docs and the old behaviour we pass three parameters to the user supplied checkIgnore function. The new library broke that. This brings the old behaviour back.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -289,8 +289,15 @@ class Config
         if ($this->shouldSuppress()) {
             return true;
         }
-        if (isset($this->checkIgnore) && call_user_func($this->checkIgnore, $isUncaught, $toLog, $payload)) {
-            return true;
+        if (isset($this->checkIgnore)) {
+            try {
+                if (call_user_func($this->checkIgnore, $isUncaught, $toLog, $payload)) {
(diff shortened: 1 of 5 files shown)


bretto36:
How long does a merge normally take? Really need this before i pay for rollbar. We occasionally have a DB server that fails and it generally uses up our limit in a day if someone isn't fast enough to fix it. Want to go the paid model but don't want to waste my money

rokob:
I will get it out today

bretto36:
Thanks @rokob

PR #154, opened 2017-05-08
Added method hinting for levels

Alfalfamale:
My PhPStorm did not agree with non existent methods

### DIFF
--- a/src/Payload/Level.php
+++ b/src/Payload/Level.php
@@ -1,5 +1,14 @@
 <?php namespace Rollbar\Payload;
 
+/**
+ * @method static critical
+ * @method static error
+ * @method static warning
+ * @method static info
+ * @method static debug
+ * @method static ignored
+ * @method static ignore
+ */
 class Level implements \JsonSerializable
 {
     private static $values;


bradacjan:
I was preparing to post a PR request containing this, when I saw you already managed to do this.

This solution provides even more information to PhpStorm. Consider this:
`/**
 * TODO ROLLBAR PR - added the dynamic properties documentation.
 * @method static Level critical
 * @method static Level error
 * @method static Level warning
 * @method static Level info
 * @method static Level debug
 * @method static Level ignored
 * @method static Level ignore
 */`

bradacjan:
Thank you for your reply. 
I find that decision a little bit unfortunate - I think that an enumeration (set of objects accessible through static methods, which is as close to an enum as you can get in PHP) is actually better suited for this than a set of constants. 
But I guess you have good reasons for that decision.

ArturMoczulski:
@bradacjan I would love to hear why you think a class with static methods is better suited for this.

Using static methods for indicating log level was forcing hardcoded dependencies inside the codebase. It was affecting potential testibility and is also a common anti-pattern. Additionally, Level as an object is useful only inside the library and there is no needs to expose more complexity than is needed. Now Level objects are still being utilized but only inside the library and are being created by a LevelFactory which is an injectable dependency.

Here is a short summary of benefits of this approach:
* hide complexity of `Rollbar\Payload\Level` from end users
* creation method of `Rollbar\Payload\Level` is injectable
* less usage of static method calls anti-pattern
* code is simpler and more obvious through a list of constants rather than a `__callStatic` mechanism

bradacjan:
Oh. The explanation that you have provided gives very strong arguments for the switch towards using constants. 

From other PRs and Issues here on Github I have understood that Rollbar Level might get in sync with Monolog levels, which I would see as a great solution (moving towards a more "standardized" solution). 

But back to your comment: all in all your proposed solution definitely sounds great (using Level objects only inside the library, exposing only well defined API, ...). 

But on a theoretical level I have a different opinion on some of your points. Because the current implementation of Level is, in fact, an enumeration. Sort of "Java style" enum. Or as close to an enumeration as you can get in PHP. 
And the static methods you mention as being an antipattern are in fact not "typical" static methods, but only very simple accessors to the corresponding enumeration instances. Therefore I see them no more as a static method antipattern than I would see, for example, Fruits::APPLE call in Java. 
In principle they are constants. Just a bit different, since they are accessed via a static method - not a static member. 

And in this context I see nothing wrong with the __callStatic method. It can be used to produce really bad code, which is why people tend to stay away from it. But if used smart, it can in fact help build great code. Look at Larvel, which uses simillar patterns a lot (Facades for example), but they are paired with other concepts (powerful IoC container in Laravel, enumeration in our case) and therefore produce code that a lot of people seem to enjoy. 

I am not suggesting that you should change your approach to how you are going to modify the Level class. I am just answering why I think that a class with static methods can be good :) .

PR #158, opened 2017-05-09
Fix infinite loop when exception has previous set

vilius-g:
When reporting exception that has `previous` property set to some value, it seem that there is an infinite loop when rollbar tries to iterate down the exception tree.


src/DataBuilder.php
```diff
@@ -339,7 +339,7 @@ protected function getExceptionTrace($exc)
         $baseException = $this->getBaseException();
         while ($previous instanceof $baseException) {
             $chain[] = $this->makeTrace($previous);
-            $previous = $exc->getPrevious();
+            $previous = $previous->getPrevious();
```
rokob: Good catch

Issue #156, opened 2017-05-09
the config item use_error_reporting is not used in code


p4ul:
**use_error_reporting**
Sets whether to respect current `error_reporting()` level or not
Default: false

If you search the code this is no longer used. Not sure if this is deliberate?

ArturMoczulski:
I used the PR from vilius-g https://github.com/rollbar/rollbar-php/pull/159 as base for my PR with added test coverage.

This is ready for review and merging.

PR #159, opened 2017-05-09
Fixed use_error_reporting regression

vilius-g:
During the refactor to 1.0 `use_error_reporting` setting
implementation was removed and it was not implemented again.

Related to https://github.com/rollbar/rollbar-php/issues/156


rokob:
Looks good, it would be nice to have a test for this as well.

PR #163, opened 2017-05-12
Centralized Log Aggregation with fluentd

olafnorge:
- added `fluent/logger` as composer dependency
- implemented `FluentSender`
- adapted `Config::class` to reflect fluent settings
- adapted README to reflect fluent sender feature

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
 
   "require": {
     "ext-curl": "*",
-    "psr/log": "~1.0.2"
+    "psr/log": "~1.0.2",
+    "fluent/logger": "^1.0"
   },
 
   "require-dev": {
--- a/src/Config.php
+++ b/src/Config.php
@@ -1,7 +1,7 @@
 <?php namespace Rollbar;
 
-use Rollbar\Payload\Payload;
 use Rollbar\Payload\Level;
+use Rollbar\Payload\Payload;
 
 if (!defined('ROLLBAR_INCLUDED_ERRNO_BITMASK')) {
     define(
@@ -93,7 +93,7 @@ class Config
     protected function updateConfig($c)
     {
         $this->configArray = $c;
-        
+
         $this->setAccessToken($c);
         $this->setDataBuilder($c);
         $this->setTransformer($c);
@@ -188,6 +188,24 @@ class Config
                 );
             }
         }
+
+        // set options for fluent sender
+        if (isset($c['handler']) && $c['handler'] == 'fluent') {
+            $default = "Rollbar\Senders\FluentSender";
+
+            if (isset($c['fluent_host'])) {
+                $c['senderOptions']['fluentHost'] = $c['fluent_host'];
+            }
+
+            if (isset($c['fluent_port'])) {
+                $c['senderOptions']['fluentPort'] = $c['fluent_port'];
+            }
+
+            if (isset($c['fluent_tag'])) {
+                $c['senderOptions']['fluentTag'] = $c['fluent_tag'];
+            }
+        }
+
         $this->setupWithOptions($c, "sender", $expected, $default);
     }
 
@@ -245,6 +263,7 @@ class Config
         $defaultClass = null,
         $passWholeConfig = false
     ) {
+
         $$keyName = isset($c[$keyName]) ? $c[$keyName] : null;
 
         if (is_null($defaultClass) && is_null($$keyName)) {
@@ -274,7 +293,7 @@ class Config
     {
         return $this->dataBuilder->makeData($level, $toLog, $context);
     }
-    
+
     public function getDataBuilder()
     {
         return $this->dataBuilder;
--- a/README.md
+++ b/README.md
@@ -207,6 +207,24 @@ $config = array(
 
 You'll also need to run the agent. See the [rollbar-agent docs](https://github.com/rollbar/rollbar-agent) for setup instructions.
 
+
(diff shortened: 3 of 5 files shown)


olafnorge:
I would love to fix the `codeclimate` issues but I have no idea what I should change. I took the `AgentSender` as reference implementation.

May you give me a hint what I can change to pass the `codeclimate` tests as well?

olafnorge:
@ArturMoczulski as I learned from the [PR](https://github.com/jenssegers/laravel-rollbar/pull/85#issuecomment-301259733) at Jens Laravel impl you are the main developer behind the rewrite/refactoring of the rollbar-php client I would like to ask you, if you can take a closer look on this PR.

I already impl. a [rollbar-fluentd forwarder](https://github.com/olafnorge/fluent-plugin-rollbar) which is also linked in the [docs](https://rollbar.com/docs/) of rollbar.

As I am using docker with the native fluentd log driver it would be awesome to get the fluent sender also included in the rollbar-php package so that I don't have to print the errors to stdout anymore. Because the errors with stack traces are often to big to get consumed in one chunk as php has limited buffer for stdout and stderr.

src/Config.php
```diff
@@ -187,21 +187,15 @@ private function setSender($c)
             $default = "Rollbar\Senders\FluentSender";
 
             if (isset($c['fluent_host'])) {
-                $c['senderOptions'] = array(
-                    'fluentHost' => $c['fluent_host']
-                );
+                $c['senderOptions']['fluentHost'] = $c['fluent_host'];
```
ArturMoczulski: The fix was correct. We use `array(...)` syntax for PHP 5.3 support. Do you think you can revert back to that?

src/Config.php
```diff
@@ -187,21 +187,15 @@ private function setSender($c)
...
             }
 
             if (isset($c['fluent_port'])) {
-                $c['senderOptions'] = array(
-                    'fluentPort' => $c['fluent_port']
-                );
+                $c['senderOptions']['fluentPort'] = $c['fluent_port'];
```
ArturMoczulski: The fix was correct. We use `array(...)` syntax for PHP 5.3 support. Do you think you can revert back to that?

src/Config.php
```diff
@@ -187,21 +187,15 @@ private function setSender($c)
...
             }
 
             if (isset($c['fluent_tag'])) {
-                $c['senderOptions'] = array(
-                    'fluentTag' => $c['fluent_tag']
-                );
+                $c['senderOptions']['fluentTag'] = $c['fluent_tag'];
```
ArturMoczulski: The fix was correct. We use `array(...)` syntax for PHP 5.3 support. Do you think you can revert back to that?

ArturMoczulski:
@olafnorge This looks great. Thanks for your efforts. Once you fix the array syntax I mentioned I will look into getting this into master.

src/Config.php
olafnorge: Replying to all above: No, because it would override the sender config. The current impl extends each of the fields by another setting.

olafnorge: It just takes the value from the corresponding key. :)

olafnorge: So, what I mean is that:

```
$c['senderOptions'] = array(
     'fluentTag' => $c['fluent_tag']
);
```
Would override the entire sender config and the last setting wins. If you set more than one additional thing like `host` and `tag` the sender config would contain only `tag`.

While 
```
$c['senderOptions']['fluentTag'] = $c['fluent_tag'];
```
would extend the sender options by whatever is set on the right side of the `=`.

I hope this makes it clearer. :)

olafnorge: Same could be achieved by an `array_merge` but why doing an additional function call as the current impl is supported since ever I would say. :)

ArturMoczulski: Oh I see. I misread this code. I though it was just usage of the [] syntax for array definition for some reason. Yeah, then this is good to go. I'll submit a github issue for this pull request.

ArturMoczulski:
@rokob @coryvirok @brianr this work prepared by @olafnorge looks great. I reviewed it and I'm opting for merging this into rollbar/rollbar-php:master

olafnorge:
Should I squash the commit?

ArturMoczulski:
@olafnorge sure 👍

olafnorge:
Hi @ArturMoczulski any updates if the PR is going to be merged?

ArturMoczulski:
@olafnorge most likely yes, although it might take a minute. unless you're waiting or are dependent on having this merge, then we will prioritize. Do you need it merged asap?

olafnorge:
Currently I am using my fork to use it in production but keeping track with all the other changes is quite painful. Would be awesome to get it asap if possible but I definitely see your need to play it safe as it is a lib used by many of your customers.

composer.json
```diff
@@ -32,7 +32,8 @@
 
   "require": {
     "ext-curl": "*",
-    "psr/log": "~1.0.2"
+    "psr/log": "~1.0.2",
+    "fluent/logger": "^1.0"
```
rokob: Should we require this is a dependency even for those that do not want to use it? Maybe we should put it in `suggest` and document that one needs to ensure they have it installed in order to use this?

rokob:
I'll go ahead and merge this as it looks good to me as well. We can deal with dependencies later.

olafnorge:
Thanks you for merging! :+1:

PR #165, opened 2017-05-16
Check if include error code context config value explicitly is not null. Error code context should be off by default.

boboldehampsink:
When `include_error_code_context` is set to "false", the `$fromConfig != null` check will falsely return `false`. Explicitly checking for a null value will fix this.

And since this is not working at the moment, include code context is now on by default while it should be off.


rokob:
I am going to merge this as we should fix this, but it is still not quite correct as it will still lead to it being on by default based on the line above which sets it to true initially. I will put up a change to move this to defaults and handle it like the other options.

Issue #38, opened 2014-11-06
Truncate long strings in large payloads to respect 128kb limit


brianr:
rollbar-gem does this: https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar.rb#L400-L410 . We should do something similar here.

ArturMoczulski:
@brianr I need a bit more explanation on what needs to be done here:

- What parts of the payload should be truncated?
- The 128kb limit is a limit for the whole payload or single field of the payload?
- Can you provide an example of when such truncating should be done?

ArturMoczulski:
Also, the rollbar-gem example you provided seems to not be up to date (lines 400-410 are not there anymore). Can you point me to where this resides now?

coryvirok:
Here's a more up-to-date link - https://github.com/rollbar/rollbar-gem/tree/master/lib/rollbar/truncation

cc @rokob for direction on how to proceed with truncation.

rokob:
The API will reject requests that are too large: https://rollbar.com/docs/api/items_post/#response-codes

> 413 | Request entity too large | Max payload size is 128kb. Try removing or truncating unnecessary large data included in the payload, like whole binary files or long strings.

The docs say 128kb, the reality is that the entire entity-body of the post request must be below 1MB. That is, the Content-Length as defined by: "Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs" must be less than 1MB.

What this means in practice is that if the payload itself is too large, we need to truncate some things to get it below the desired threshold. We want to truncate strings that are values in the request body (i.e. not keys in a JSON object).

The way the gem decides to do this, is to pick a max payload size of 0.5 MB (https://github.com/rollbar/rollbar-gem/blob/bf6d439be15455b557886fc73304ba00303aa278/lib/rollbar/truncation.rb#L12), and then to apply strategies in order until the payload satisfies this size limit: https://github.com/rollbar/rollbar-gem/blob/bf6d439be15455b557886fc73304ba00303aa278/lib/rollbar/truncation.rb#L22-L25

Those strategies https://github.com/rollbar/rollbar-gem/tree/bf6d439be15455b557886fc73304ba00303aa278/lib/rollbar/truncation start with some conservative truncations such as removing the middle of the stack trace if there are an excessive number of frames, and get more aggressive as the payload continues to be too large.

There is a bit of flexibility in how to do the truncation, but the main rules are:

[...]

PR #167, opened 2017-05-16
GitHub Issue #38: truncate payload

ArturMoczulski:
Truncation is ready for review. This was implemented to resemble truncation strategies from the Ruby notifier as close as possible.


tests/Truncation/FramesStrategyTest.php
```diff
@@ -0,0 +1,70 @@
...
+     * @dataProvider executeProvider
+     */
+    public function testExecute($data, $expected)
+    {
+        $dataBuilder = new DataBuilder(array(
+            'accessToken' => 'abcd1234efef5678abcd1234567890be',
+            'environment' => 'tests'
+        ));
+                    
+        $strategy = new FramesStrategy($databuilder);
```
rokob: s/databuilder/dataBuilder/

rokob:
This looks very good, I'll give it another once over to see if we can get this merged today.

src/Truncation/FramesStrategy.php
```diff
@@ -0,0 +1,32 @@
+<?php namespace Rollbar\Truncation;
+
+class FramesStrategy extends AbstractStrategy
+{
+    
+    const FRAMES_OPTIMIZATION_RANGE = 150;
```
rokob: This probably comes from matching what the Gem does, but I feel like it is really large. Maybe we can make this 75 instead so that we get 150 total frames rather than 300 when we truncate.

ArturMoczulski: 👍

rokob:
Tests pass, but lint doesn't like the long lines in the tests. Should be a simple fix then we can merge this and release 1.1

ArturMoczulski:
Okay, this is ready for merging

rokob:
👍

PR #169, opened 2017-05-16
Move packfire/php5.3-compat from require to suggest in composer.json

### LINKED ISSUE #162: Remove hard dependency on packfire/php5.3-compat
At this point, PHP 5.3 accounts for less than 1% of all Composer users ([source](https://seld.be/notes/php-versions-stats-2017-1-edition)). The README for this package notes that Composer is the preferred installation method ([source](https://github.com/rollbar/rollbar-php#using-composer-recommended)).

Given this, I believe it would be reasonable to shift from [requiring](https://github.com/rollbar/rollbar-php/blob/1b48a62b86d9c918dcf62ed2a96ca8d06e8eb279/composer.json#L36) the `packfire/php5.3-compat` package to [suggesting](https://getcomposer.org/doc/04-schema.md#suggest) it for PHP 5.3 users. It would make for one less package that the 99% using PHP 5.4 and above have to download and bundle in our builds to be able to use Rollbar.

I'd be happy to submit a PR for this change if it seems reasonable to the maintainers as well.

elazar:
Fixes #162

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,8 @@
     "mockery/mockery": "0.9.*",
     "squizlabs/php_codesniffer": "2.*",
     "codeclimate/php-test-reporter": "dev-master",
-    "monolog/monolog": "*"
+    "monolog/monolog": "*",
+    "packfire/php5.3-compat": "*"
   },
 
   "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -32,8 +32,7 @@
 
   "require": {
     "ext-curl": "*",
-    "psr/log": "~1.0.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -44,6 +43,10 @@
     "monolog/monolog": "*"
```
rokob: I think adding `"packfire/php5.3-compat": "*"` to `require-dev` will get the tests to pass and will still satisfy the desire not to install it for everyone.

elazar: Yup, that did it. Thanks.

rokob:
👍

PR #172, opened 2017-05-19
Testing for existance of `exec()`

hellopablo:
Tests for existence of `exec()` function so as not to raise notices on systems where "exec() has been disabled for security reasons".

### DIFF
--- a/src/Defaults.php
+++ b/src/Defaults.php
@@ -19,9 +19,11 @@ class Defaults
     private static function getGitHash()
     {
         try {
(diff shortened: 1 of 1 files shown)


rokob:
LGTM

rokob:
I took me some digging to confirm that `function_exists` actually returns false for disabled functions, because there are a lot of things online saying that it will return true but just error out. It seems that as of at least 5.3+ if a function is disabled via the ini file then this method will work. I think that is good enough.

PR #175, opened 2017-05-22
 Don't ignore native PHP error handler

pavel-krulov:
Don't ignore native PHP error handler, when rollbar error handler setup.

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -76,11 +76,12 @@ class Rollbar
 
     public static function errorHandler($errno, $errstr, $errfile, $errline)
     {
-        if (is_null(self::$logger)) {
-            return;
+        if (null !== self::$logger) {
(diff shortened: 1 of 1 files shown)


coryvirok:
Good catch, thanks. @ArturMoczulski can you take a look? We will also want to make sure there's a test for this case.

rokob:
This looks good to me, this was part of the original set of changes moving from 0.18.2 to 1.0.0, c.f. https://github.com/rollbar/rollbar-php/commit/709fae9ee7f389b48e762ae8b5ac41b3cbbef635#diff-b156308dfc58bccc36efd0ce1cb4d496L68

The behaviour in 0.18.2 is the same as this PR, so for compatibility alone this is a good change.

ArturMoczulski:
Let's merge this PR so I can start working on a test

rokob:
👍

Issue #135, opened 2017-04-26
DataBuilder::getUrl() expects HTTP_X_FORWARDED_PROTO breaks with multiple forwarded protocols


amoskyler:
The following method only expects a single value in `HTTP_X_FORWARDED_PROTO`, which is inconsistent with the handling of the `HTTP_X_FORWARDED_FOR` header, which delimits on ',' and pops the last host.

The following code:
```
        if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
            $proto = strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']);
        } elseif (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') {
            $proto = 'https';
        } else {
            $proto = 'http';
        }
```

Should be modified as follows:
```
$parts = explode(',', $_SERVER['HTTP_X_FORWARDED_PROTO'])
$proto = strtolower($parts[0]);
```

rokob:
HTTP_X_FORWARDED_PROTO really should be single valued, but you are right as it is not actually standardized it can be the same as FOWARDED_FOR

ArturMoczulski:
This is ready in PR https://github.com/rollbar/rollbar-php/pull/179

PR #179, opened 2017-05-25
GitHub Issue #135: HTTP_X_FORWARDED_PROTO  comma separated

ArturMoczulski:


rokob:
In PHP 5.3 you cannot use the array index syntax with the result of a function call, need to use a temp variable.

ArturMoczulski:
Good catch @rokob

rokob:
👍

Issue #177, opened 2017-05-25
PHP 5.4 issue with php://input


p4ul:
This function breaks functionality in our script, as it clears out PUT variable body so nothing is returned when our script tries to subsequently access it.

```
    protected function setRequestBody($config)
    {
        $this->requestBody = $this->tryGet($config, 'requestBody');
        
        if (!$this->requestBody) {
            $this->requestBody = file_get_contents("php://input");
        }
    }
```

As file_get_contents("php://input"); can only be called once in older versions of PHP.

See:
Note: Prior to PHP 5.6, a stream opened with php://input could only be read once; the stream did not support seek operations. However, depending on the SAPI implementation, it may be possible to open another php://input stream and restart reading. This is only possible if the request body data has been saved. Typically, this is the case for POST requests, but not other request methods, such as PUT or PROPFIND

ArturMoczulski:
Let me calrify this. In your script you're also trying to access php://input, but now it's empty since it has been already read by Rollbar SDK. Is that correct?

p4ul:
@ArturMoczulski yes

ArturMoczulski:
Solution for this is waiting to be merged https://github.com/rollbar/rollbar-php/pull/181

It's not perfect, but I don't think we can do anything more for PHP < 5.6. This PR disables reading php://input by default and allows to enable it through `include_raw_request_body` config option. It also saves the content of the stream in `$_SERVER['php://input']` in case you still want to access it.

I hope this solves your problem @p4ul

PR #181, opened 2017-05-26
GitHub Issue #177: make reading from php://input configurable

ArturMoczulski:
In response to feedback from #177 I decided to make reading the raw request body from php://input disabled by default and configurable through `include_raw_request_body`.

Unfortunately, I don't see any way to overcome PHP < 5.6 limitation of single read from php://input so I combined making it configurable and also saving the content of the stream for later use in `$_SERVER['php://input']`. This is not perfect, but I don't think we can do anything more for < 5.6 users due to PHP limitations.


src/DataBuilder.php
```diff
@@ -160,6 +162,12 @@ protected function setSendMessageTrace($config)
         $fromConfig = $this->tryGet($config, 'send_message_trace');
         $this->sendMessageTrace = self::$defaults->sendMessageTrace($fromConfig);
     }
+    
+    protected function setIncludeRawRequestBody($config)
+    {
+        $fromConfig = $this->tryGet($config, 'include_raW_request_body');
```
rokob: s/include_raW_request_body/include_raw_request_body/

README.md
```diff
@@ -460,6 +460,17 @@ Default: No proxy
...
 
+<dt>include_raw_request_body</dt>
+<dd>Include the raw request body from php://input in payloads.
+Note: in PHP < 5.6 if you enable this, php://input will be empty for PUT requests
+as Rollbar SDK will read from it 
+@see http://php.net/manual/pl/wrappers.php.php#refsect2-wrappers.php-unknown-unknown-descriptiop
+If you still want to read the request body for your PUT requests Rollbar SDK saves
+the content of php://input in $_SERVER['php://input']
+
+Default: `false`
```
rokob: Is this something that most people assume should be there though? I feel like the default should be what we have been doing so far and allow people using < 5.6 to deal with the problem. But I am not sure, it is a shitty situation that I don't see a better solution for.

ArturMoczulski: This alters the default behavior of PHP in what I would call an unexpected place. Because of this I would opt for keeping this disabled by default.

src/DataBuilder.php
```diff
@@ -193,10 +201,12 @@ protected function setRequestParams($config)
 
     protected function setRequestBody($config)
     {
+        
         $this->requestBody = $this->tryGet($config, 'requestBody');
         
-        if (!$this->requestBody) {
+        if (!$this->requestBody && $this->includeRawRequestBody) {
             $this->requestBody = file_get_contents("php://input");
+            $_SERVER['php://input'] = $this->requestBody;
```
ftrrtf: Saving request body into `$_SERVER` make sense to keep _only_ for old PHP versions due to performance reason. And even withing old PHP versions it's a huge hack, IMO, to build the logic of own application based on some workaround behavior of vendor logging library. I would recommend to just remove this line.

ArturMoczulski: @rokob For the same reasons I opt in for `include_raw_request_body` to be disabled by default.

rokob: What if we did
```php
if (version_compare(PHP_VERSION, '5.6.0') < 0) {
    $_SERVER['php://input'] = $this->requestBody;
}
```

At least then we only do this hack when we need it.

README.md
```diff
@@ -517,6 +517,15 @@ Default: No proxy
...
 
+
+<dt>include_raw_request_body</dt>
+<dd>Include the raw request body from php://input in payloads.
+Note: in PHP < 5.6 if you enable this, php://input will be empty for PUT requests
+as Rollbar SDK will read from it 
+@see http://php.net/manual/pl/wrappers.php.php#refsect2-wrappers.php-unknown-unknown-descriptiop
+If you still want to read the request body for your PUT requests Rollbar SDK saves
+the content of php://input in $_SERVER['php://input']
+
```
rokob: Let's mention the default being false for this

src/DataBuilder.php
```diff
@@ -209,10 +217,12 @@ protected function setRequestParams($config)
 
     protected function setRequestBody($config)
     {
+        
         $this->requestBody = $this->tryGet($config, 'requestBody');
         
-        if (!$this->requestBody) {
+        if (!$this->requestBody && $this->rawRequestBody) {
             $this->requestBody = file_get_contents("php://input");
+            $_SERVER['php://input'] = $this->requestBody;
```
rokob: Let's do something like to only do this when absolutely necessary

```php
if (version_compare(PHP_VERSION, '5.6.0') < 0) {
    $_SERVER['php://input'] = $this->requestBody;
}
```

src/Defaults.php
```diff
@@ -99,6 +99,11 @@ public function localVarsDump($localVarsDump = null)
         return $localVarsDump !== null ? $localVarsDump : $this->defaultLocalVarsDump;
     }
 
+    public function rawRequestBody($rawRequestBody = null)
+    {
+        return $rawRequestBody ? $rawRequestBody : $this->defaultRawRequestBody;
```
rokob: We should be testing for `$rawRequestBody !== null`

tests/DataBuilderTest.php
```diff
@@ -982,12 +982,14 @@ public function testSetRequestBody()
...
             'accessToken' => 'abcd1234efef5678abcd1234567890be',
-            'environment' => 'tests'
+            'environment' => 'tests',
+            'include_raw_request_body' => true
         ));
         $output = $dataBuilder->makeData(Level::fromName('error'), "testing", array());
         $requestBody = $output->getRequest()->getBody();
         
         $this->assertEquals($streamInput, $requestBody);
+        $this->assertEquals($streamInput, $_SERVER['php://input']);
```
rokob: Only do this check on less than 5.6

rokob:
Other than the comments this looks good

rokob:
This LGTM but there is still an issue with the tests taking too long, I'm going to merge this and deal with the test timeouts elsewhere.

PR #183, opened 2017-05-28
Fixed crash when getting stack trace from file with unexpected line endings.

bradacjan:
I.e., for file with line endings different from system default (the PHP_EOL). 

For example: if a method in the stack trace was contained in a file with Unix
line endings("\n") on a development Windows machine (where PHP_EOL is "\r\n").
Then the application crashes endlessly, as a new exception is thrown every time
a stack trace is being generated for the last one. Because the program tries
to read line X from an array with only one element (since the split of file
into lines by line breaks failed) which in turn throws another error.

The error was quite severe as it loops on itself endlessly.

### DIFF
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -424,12 +424,8 @@ class DataBuilder implements DataBuilderInterface
             return;
         }
 
-        $source = explode(PHP_EOL, file_get_contents($filename));
-        if (!is_array($source)) {
-            return;
-        }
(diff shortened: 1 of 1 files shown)


bradacjan:
I have accidentally used a wrong email address for this pull request. My apologies.

bradacjan:
I made mistakes in the commit message. My appologies, will fix and post a new PR.

PR #184, opened 2017-05-29
Prevent endless loop if the $exc->getPrevious() equals the exception itself.

bradacjan:
Some external libraries in our project generate exception objects which contain reference to themselves as the previous exception.

This commit allows the Rollbar code to detect such a situation and break out from an otherwise endless loop.


rokob:
LGTM

rokob:
Ah shit I shouldn't have merged this, this is wrong, I'll fix it.

rokob:
Fixed in https://github.com/rollbar/rollbar-php/commit/d4c16311951d5a4c13f2b21f88c320cf6c6cd210

PR #185, opened 2017-05-29
Fixed crash when getting stack trace from file with unexpected line endings (different from system deafult, i.e. PHP_EOL)

bradacjan:
For example: if a method in the stack trace was contained in a file with Unix
line endings("\n") on a development Windows machine (where PHP_EOL is "\r\n").
Then the application crashes endlessly, as a new exception is thrown every time
a stack trace is being generated for the last one. Because the program tries
to read line X which means an element on a position X-1 from an array with only 
one element (since the split of the file into lines by line breaks failed) which in turn 
throws a new error, since X-1 is greater than 1, i.e., out of bounds.

The error was quite severe as it loops on itself endlessly.


src/DataBuilder.php
```diff
@@ -980,4 +976,33 @@ public function needsTruncating(array $payload)
...
+            if (substr_count($rawSource, "\n") > substr_count($rawSource, "\r")) {
+                $source = explode("\n", $rawSource);
+                return $source;
+            }
+            
+            $source = explode("\r", $rawSource);
+            return $source;
+        }
+
+        $source = str_replace(array("\n", "\t", "\r"), '', $source);
```
rokob: Don't we always want to do this in the end?

bradacjan: Yes. You are right, this is inconsistent (we returned from the if branches without replacing the linebreaks). My mistake. I have fixed it in the next commit.

Issue #137, opened 2017-04-27
Add helper function to inject javascript snippet in views


rokob:
In the rollbar gem we provide the ability to inject the snippet for rollbarjs into your html rendered views. We should provide some mechanism that makes that easy to do in php as well.

PR #186, opened 2017-05-29
Github Issue #137: Add helper function to inject javascript snippet in views

ArturMoczulski:
Implementation of injecting RollbarJS snippet through PHP SDK similar to the Rollbar gem #137 

This is still work in progress.


ArturMoczulski:
Ready for review

src/RollbarJsHelper.php
```diff
@@ -0,0 +1,185 @@
...
+     * @param array $headers Response headers usually retrieved through
+     * headers_list() used to verify if nonce should be added to script
+     * tags based on Content-Security-Policy
+     * @param string $nonce Content-Security-Policy nonce string if exists
+     *
+     * @return string
+     */
+    public function addJs($headers = null, $nonce = null)
+    {
+        return
```
rokob: So this returns one string with two script tags, I think some browsers can execute script tags in arbitrary order so we suggest in the rollbar.js docs putting both the _rollbarConfig variable definition and the snippet inside one script tag. For most browsers it doesn't matter as things are executed top to bottom, but to be safe we should generate one script tag which has both the config var and the snippet.

ArturMoczulski: Ok, will squash this in one script tag. Althoug, I'm pretty sure that is not how rollbar-gem does this: 
https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/middleware/js.rb#L85
https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/middleware/js.rb#L105-L111

We might fix this in rollbar-gem as well then.

rokob:
Nice

Issue #153, opened 2017-05-05
Issues with using a custom API endpoint


brianr:
I ran into a few problems using 1.0.1 with a custom endpoint (i.e. sending data to a sandbox instance of Rollbar instead of production):

- I expected the config variable name to be "endpoint", like it is in most/all other notifiers. It's actually `base_api_url`.
  - _Suggestion: add support for `endpoint`, and deprecate `base_api_url`_
- After switching to `base_api_url`, I expected the URL format to be `https://mydomain.com/api/1/`, just like the documented default (`https://api.rollbar.com/api/1/`). However, this did not work - requests via the CurlSender were 404ing. To fix it, I had to change it to `https://mydomian.com/api/1/item/`.
  -  _Suggestion: `endpoint` should expect a url like `https://mydomain.com/api/1/`, not `https://mydomain.com/api/1/item/`_
  - _Second suggestion: research the behavior of base_api_url in rollbar-php 0.18 and ensure that base_api_url works the same way in 1.0.x_

rokob:
Changing `base_api_url` to `endpoint` is a good idea.

Removing `/item/` from the endpoint is also a good idea.

This is not how it used to work, so this should be fixed quickly for backward compatibility

```php
public $base_api_url = 'https://api.rollbar.com/api/1/'
```

ArturMoczulski:
I will submit a PR for this soon.

ArturMoczulski:
Ready for review in https://github.com/rollbar/rollbar-php/pull/189

rokob:
Merged

PR #189, opened 2017-05-30
GitHub Issue #153: Issues with using a custom API endpoint

ArturMoczulski:
In 1.0.0 the behavior of `base_api_url` was changed and required additional _item/_ at the end. This was both not in sync with other SDKs and backwards incompatible.

This PR brings back the the behavior from 0.18.2 which doesn't require the _item/_ URL path.

Also, `base_api_url` is deprecated from now on in favor of `endpoint` which is in sync with how other SDKs work.

### DIFF
--- a/tests/ConfigTest.php
+++ b/tests/ConfigTest.php
@@ -207,6 +207,40 @@ class ConfigTest extends \PHPUnit_Framework_TestCase
         ));
         $c->send($p, $this->token);
     }
+    
+    public function testEndpoint()
+    {
+        $payload = m::mock("Rollbar\Payload\Payload");
+            
+        $config = new Config(array(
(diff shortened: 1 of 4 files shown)


tests/ConfigTest.php
```diff
@@ -208,6 +208,38 @@ public function testSender()
         $c->send($p, $this->token);
     }
     
+    public function testEndpoint()
```
rokob: Can we add a test that doesn't set the endpoint in the config just to ensure the default is propagating through properly?

rokob:
👍

tests/ConfigTest.php
ArturMoczulski: Done

PR #190, opened 2017-05-30
Decoupling of RollbarLogger in init

jasny:
Allow passing a `RollbarLogger` in init instead of creating one from config.
This allows to use a custom `RollbarLogger` by extending the class. You can also get the logger from a container for container based frameworks.

### DIFF
--- a/tests/RollbarTest.php
+++ b/tests/RollbarTest.php
@@ -7,6 +7,11 @@ if (!defined('ROLLBAR_TEST_TOKEN')) {
 use Rollbar\Rollbar;
 use Rollbar\Payload\Level;
 
+/**
+ * Usage of static method Rollbar::logger() is intended here.
+ *
+ * @SuppressWarnings(PHPMD.StaticAccess)
+ */
 class RollbarTest extends \PHPUnit_Framework_TestCase
 {
 
--- a/tests/RollbarTest.php
+++ b/tests/RollbarTest.php
@@ -15,13 +15,56 @@ class RollbarTest extends \PHPUnit_Framework_TestCase
         'environment' => 'test'
     );
     
-    public function setUp()
+    public function tearDown()
     {
+        $reflLoggerProperty = new \ReflectionProperty(Rollbar::class, 'logger');
+        $reflLoggerProperty->setAccessible(true);
+        $reflLoggerProperty->setValue(null);
+    }
+    
+    public function testInitWithConfig()
+    {
+        Rollbar::init(self::$simpleConfig);
+        
+        $this->assertInstanceOf(RollbarLogger::class, Rollbar::logger());
+        $this->assertAttributeEquals(new Config(self::$simpleConfig), 'config', Rollbar::logger());
+    }
+    
+    public function testInitWithLogger()
+    {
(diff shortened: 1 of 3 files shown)


ArturMoczulski:
This is great. Do you think you can add a test case for this?

ftrrtf:
Cool idea, but naming can be better. `$config` is not just config anymore.

jasny:
@ArturMoczulski Yest, but not easily. These are static methods altering the global state. We need to reset the state in `RollbarTest::tearDown()`.
@ftrrtf We could name it `$logger`. Any other ideas?

I also noticed that it's going to break if you pass a `RollbarLogger` when `self::$logger` is already set. What do you think should happen? We could simply replace it, but that might lead to unexpected behavior. Alternatively an exception can be through

```
throw new BadMethodCallException("It's not possible to replace the global rollbar logger. You may any specify a new configuration.");
```

Please let me know and I'll make the modifications.

ArturMoczulski:
Yes, but we will need it for test coverage. The extent of the test can be as minimal as just checking if subsequent call to `Rollbar::logger()` returns the same instance as the one passed in `Rollbar::init()`.

Passing a new instance of RollbarLogger when one already exists in `self::$logger` should replace the old one. I can't think of any reasons why that would be bad.

rokob:
L30 needs to change as you have mentioned, it should be the same as L21, which makes me think it should delegate to a shared method like `setLogger`. The `config` variable shouldn't be named `logger` either because it is a config or a logger, so maybe just `configOrLogger` which is at least explicit.

jasny:
@rokob Replacing the logger when passing a new configuration in `Rollbar::init()` would break backwards compatibility.

tests/RollbarTest.php
```diff
@@ -14,14 +14,67 @@ class RollbarTest extends \PHPUnit_Framework_TestCase
...
+    public function tearDown()
+    {
+        self::clearLogger();
+    }
+    
+    public function testInitWithConfig()
+    {
+        Rollbar::init(self::$simpleConfig);
+        
+        $this->assertInstanceOf('Rollbar\RollbarLogger', Rollbar::logger());
```
ArturMoczulski: Add this comment above the method signature to suppress PHPMD issues in CodacY:

```php
    /**
     * Usage of static method Rollbar::logger() is intended here.
     *
     * @SuppressWarnings(PHPMD.StaticAccess)
     */
```

tests/RollbarTest.php
```diff
@@ -14,14 +14,67 @@ class RollbarTest extends \PHPUnit_Framework_TestCase
...
+        $this->assertAttributeEquals(new Config(self::$simpleConfig), 'config', Rollbar::logger());
+    }
+    
+    public function testInitWithLogger()
+    {
+        $logger = $this->getMockBuilder('Rollbar\RollbarLogger')->disableOriginalConstructor()->getMock();
+
+        Rollbar::init($logger);
```
ArturMoczulski: Add this comment above method signature to suppress PHPMD issues in Codacy:

```php
    /**
     * Usage of static method Rollbar::logger() is intended here.
     *
     * @SuppressWarnings(PHPMD.StaticAccess)
     */
```

PR #178, opened 2017-05-25
GitHub Issue #173: add stack frame arguments values to traces

ArturMoczulski:
This PR adds `local_vars_dump` configuration option. If enabled traces sent to Rollbar will include the values of arguments passed to each frame in stack traces.

Issue #166, opened 2017-05-16
capture_error_stacktraces is not used


boboldehampsink:
There's no way to disable the use generation of stacktraces anymore.

rokob:
The option `capture_error_stacktraces` used to exist to turn this off but has gone away as of 1.0, this should be added back.

ArturMoczulski:
@rokob Not sure about this. Just went through the log and I can't see any code related to `capture_error_stacktraces` at all. There was documentation in README.md for it, but no actual code.

Either way, I'm going to implement this for 1.2.0

ArturMoczulski:
This is ready for review in PR https://github.com/rollbar/rollbar-php/pull/192

rokob:
Merged

PR #192, opened 2017-05-30
GitHub Issue #166: capture_error_stacktraces is not used

ArturMoczulski:
Implementation of `capture_error_stacktraces` in 1.2.0.

This PR is based on PR https://github.com/rollbar/rollbar-php/pull/178 to pull in changes to `generateErrorWrapper`


ArturMoczulski:
Ready for review.

src/Defaults.php
```diff
@@ -84,7 +84,19 @@ private static function getScrubFields()
     
     public function sendMessageTrace($sendMessageTrace = null)
     {
-        return $sendMessageTrace ? $sendMessageTrace : $this->defaultSendMessageTrace;
+        return $sendMessageTrace !== null ? $sendMessageTrace : $this->defaultSendMessageTrace;
```
rokob: 👍

src/Defaults.php
```diff
@@ -84,7 +84,19 @@ private static function getScrubFields()
...
+    public function captureErrorStacktraces($captureErrorStracktraces = null)
+    {
+        return $captureErrorStracktraces !== null ?
+            $captureErrorStracktraces :
+            $this->defaultCaptureErrorStacktraces;
+    }
+    
+    public function localVarsDump($localVarsDump = null)
+    {
+        return $localVarsDump ? $localVarsDump : $this->defaultLocalVarsDump;
```
rokob: Shouldn't we follow the same pattern and use
```php
return $localVarsDump !== null ? $localVarsDump : $this->defaultLocalVarsDump;
```

rokob: It doesn't particularly matter considering the default is `false`, but if we wanted to change that in the future this would matter.

src/DataBuilder.php
```diff
@@ -980,4 +1007,27 @@ public function needsTruncating(array $payload)
...
+     * @param string $errline
+     *
+     * @return ErrorWrapper
+     */
+    public function generateErrorWrapper($errno, $errstr, $errfile, $errline)
+    {
+        if ($this->captureErrorStacktraces) {
+            $backTrace = array_slice(
+                debug_backtrace($this->localVarsDump ? 0 : DEBUG_BACKTRACE_IGNORE_ARGS),
+                2
```
rokob: Is this still 2 or should it be 3? I am not sure if we added a level of indirection here.

rokob:
LGTM, a couple minor questions but this looks basically good to go.

src/Defaults.php
ArturMoczulski: I changed this. `$localVarsDump ? $localVarsDump : $this->defaultLocalVarsDump` is buggy

ArturMoczulski:
@ezarowny suggested to check out how Sentry deals with local variables in stack frames. I think it's worth a shot so I'll do some research

src/DataBuilder.php
ArturMoczulski: Just tested this. Slicing at 2 still produces correct trace depth.

PR #197, opened 2017-06-08
Remove fluent log output from composer test

olafnorge:
- isolated tests from each other
- fixed broken tests

Unfortunately I don't know how to fix this test:

```
There was 1 failure:

1) Rollbar\DataBuilderTest::testPersonFuncException
Exception in person_fn was not caught.

/var/www/tests/DataBuilderTest.php:638
```

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,7 +6,7 @@
 	 convertErrorsToExceptions="true"
 	 convertNoticesToExceptions="true"
 	 convertWarningsToExceptions="true"
-	 processIsolation="false"
+	 processIsolation="true"
 	 stopOnFailure="false"
 	 syntaxCheck="false">
   <testsuites>
--- a/tests/ReadmeTest.php
+++ b/tests/ReadmeTest.php
@@ -30,36 +30,36 @@ class ReadmeTest extends \PHPUnit_Framework_TestCase
                 'environment' => 'production'
             )
         );
-        
+
         try {
             throw new \Exception('test exception');
         } catch (\Exception $e) {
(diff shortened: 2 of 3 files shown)


olafnorge:
Seems like the process isolation breaks the code coverage report. Maybe it's sufficient to reconfigure travis to allow longer execution time.

If that is not possible/feasible the only fix I can see is that the Rollbar class needs to be refactored in a way that it saves any registered error handler so that a `reset` function can iterate over the stack and restore the error handlers until the native PHP error handler is restored.

ArturMoczulski:
@olafnorge I like process isolation more than Rollbar::reset() and BaseUnitTestCase. Those are things we might introduce in future, but not in this case. The process isolation fix is simpler and something that we should do anyway IMHO.

It looks like Travis is failing due to person_fn related test. This is not a result of your code changes, so don't worry about it. I'm working on a fix for this right now. When it's done we will sync up your PR with master and it should be passing.

ArturMoczulski:
The fix for test failure is in PR https://github.com/rollbar/rollbar-php/pull/201

Travis CI should be green after PR https://github.com/rollbar/rollbar-php/pull/201 is merged to master and then this PR is synced with master.

rokob:
Ran the tests locally and they pass on current master, so rebasing and merging...

rokob:
Okay so I merged this because the tests passed locally for me, the tests continue to pass on a subset of PHP versions, and rerunning the tests causes them to pass on a different subset of PHP versions. The error on some versions is a timeout, which makes me think we have a race condition in the tests somehow.

Issue #188, opened 2017-05-29
Remove fluent log output from composer test


ArturMoczulski:
When I run `composer test` I get a bunch of output from fluent tests not being able to connect to fluent service. This needs to be cleaned up.

olafnorge:
Hi @ArturMoczulski,

I digged a little bit around and found out that, for some reason, the fluent sender is registered as handler when `Rollbar\ReadmeTest::testQuickStart()` gets executed.

```
#0 /var/www/vendor/fluent/logger/src/FluentLogger.php(346): Fluent\Logger\FluentLogger->connect()
#1 /var/www/vendor/fluent/logger/src/FluentLogger.php(390): Fluent\Logger\FluentLogger->reconnect()
#2 /var/www/vendor/fluent/logger/src/FluentLogger.php(363): Fluent\Logger\FluentLogger->postImpl(Object(Fluent\Logger\Entity))
#3 /var/www/src/Senders/FluentSender.php(69): Fluent\Logger\FluentLogger->post('rollbar', Array)
#4 /var/www/src/Config.php(405): Rollbar\Senders\FluentSender->send(Array, 'ad865e76e7fb496...')
#5 /var/www/src/RollbarLogger.php(50): Rollbar\Config->send(Array, 'ad865e76e7fb496...')
#6 /var/www/src/Rollbar.php(69): Rollbar\RollbarLogger->log(Object(Rollbar\Payload\Level), Object(Exception), Array)
#7 /var/www/tests/ReadmeTest.php(37): Rollbar\Rollbar::log(Object(Rollbar\Payload\Level), Object(Exception))
#8 [internal function]: Rollbar\ReadmeTest->testQuickStart()
#9 /var/www/vendor/phpunit/phpunit/src/Framework/TestCase.php(908): ReflectionMethod->invokeArgs(Object(Rollbar\ReadmeTest), Array)
#10 /var/www/vendor/phpunit/phpunit/src/Framework/TestCase.php(768): PHPUnit_Framework_TestCase->runTest()
#11 /var/www/vendor/phpunit/phpunit/src/Framework/TestResult.php(612): PHPUnit_Framework_TestCase->runBare()
#12 /var/www/vendor/phpunit/phpunit/src/Framework/TestCase.php(724): PHPUnit_Framework_TestResult->run(Object(Rollbar\ReadmeTest))
[...]

PR #198, opened 2017-06-09
Remove fluent log output from composer test

olafnorge:
- introduce `reset` function in `Rollbar` class
- introduce custom base class `BaseUnitTestCase` which calls `Rollbar::reset()` in `tearDown` for each test
- all tests inherit from new `BaseUnitTestCase` class

Alternative approach for [PR 197](https://github.com/rollbar/rollbar-php/pull/197) to fix [issue 188](https://github.com/rollbar/rollbar-php/issues/188)

Unfortunately I don't know how to fix this test:

```
There was 1 failure:

1) Rollbar\DataBuilderTest::testPersonFuncException
Exception in person_fn was not caught.

/var/www/tests/DataBuilderTest.php:638
```

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -52,11 +52,11 @@ class Rollbar
     {
         set_exception_handler('Rollbar\Rollbar::exceptionHandler');
     }
-    
+
     public static function exceptionHandler($exception)
     {
         self::log(Level::error(), $exception, array(Utilities::IS_UNCAUGHT_KEY => true));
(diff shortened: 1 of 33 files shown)


ArturMoczulski:
@olafnorge this is great stuff, but in this case I like PR https://github.com/rollbar/rollbar-php/pull/197 more. Having that said, those are code changes that might become very useful in the near future so we might pull them in as part of working on something else.

Issue #140, opened 2017-04-28
Refactor usage of levels in log methods


ArturMoczulski:
@rokob has suggested that usage of Level class needs to be refactored. I completely agree. Currently it's not clear what to use and also not it's compatible with Monolog library.

ArturMoczulski:
I strongly suggest we include this in 1.1.0

PR #199, opened 2017-06-09
GitHub Issue #140: Refactor usage of levels in log methods

ArturMoczulski:
This removes most of the usage of static calls to Level class and clarifies the confusion between using `Psr\Log\Level` and `Rollbar\Payload\Level`.

https://github.com/rollbar/rollbar-php/issues/140

### DIFF
--- a/src/Payload/Level.php
+++ b/src/Payload/Level.php
@@ -1,19 +1,7 @@
 <?php namespace Rollbar\Payload;
 
-<<<<<<< HEAD
 use Rollbar\LevelFactory;
 
-=======
-/**
- * @method static Level critical()
- * @method static Level error()
- * @method static Level warning()
- * @method static Level debug()
- * @method static Level info()
- * @method static Level ignored()
- * @method static Level ignore()
- */
->>>>>>> origin/master
 class Level implements \JsonSerializable
 {
     /**
--- a/src/Payload/Level.php
+++ b/src/Payload/Level.php
@@ -2,16 +2,27 @@
 
 class Level implements \JsonSerializable
 {
-    const IGNORE = 'ignore';
-    const IGNORED = 'ignored';
-    const DEBUG = 'debug';
-    const INFO = 'info';
-    const NOTICE = 'notice';
-    const WARNING = 'warning';
-    const ERROR = 'error';
-    const CRITICAL = 'critical';
-    const ALERT = 'alert';
+    /**
+     * Those are PSR-3 compatible loggin levels. They are mapped to Rollbar
+     * service supported levels in Level::init()
+     */
     const EMERGENCY = 'emergency';  
+    const ALERT = 'alert';
+    const CRITICAL = 'critical';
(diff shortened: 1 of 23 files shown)


src/Payload/Level.php
```diff
@@ -2,6 +2,17 @@
 
 class Level implements \JsonSerializable
 {
+    const IGNORE = 'ignore';
```
coryvirok: What are these two "ignore(d)" levels for? They don't map to any level we have on the backend. Same for "alert" and "emergency".

coryvirok:
Much cleaner!

src/Payload/Level.php
ArturMoczulski: `IGNORE` and `IGNORED` are synonymous and they can be used to completely silence the logger if used in `minimumLevel` config option. I think it's mostly a leftover from 0.18.2, I don't really see the point and it's also missing from README.md. I highly doubt anybody is using them so I'm happy to remove them.

`ALERT` and `EMERGENCY` are PSR-3 levels and are mapped to Rollbar levels here: https://github.com/rollbar/rollbar-php/blob/master/src/Payload/Level.php#L16-L33

coryvirok: Aha! Thanks for the info. WOuld you mind adding a comment to this code and point to where they are mapped to the Rollbar levels?

ArturMoczulski:
This is ready for review @coryvirok @rokob

tests/LevelTest.php
```diff
@@ -7,17 +7,22 @@ class LevelTest extends \PHPUnit_Framework_TestCase
 {
     public function testLevel()
     {
-        $l = Level::TEST();
-        $this->assertNull($l);
+        try {
+            $level = Level::TEST();
+            $this->fail();
+        } catch (\Exception $exception) {
+            $this->assertTrue(true);
```
olafnorge: What is the sense behind this block? I don't what is tested here.

rokob: This appears to be testing that `Level::TEST()` throws an exception. The alternative would be to use `expectException()`

rokob:
This looks good to me, we will need to have a bump in version beyond the bugfix level because this changes the API

rokob:
Wait I am wrong about that, this uses the `__callStatic` metaprogramming magic to patch back those calls and redirect them to the constants so this is still api compatible right?

ArturMoczulski:
@rokob Yes, this is backward compatible. The static methods still work but I marked that as deprecated in favor of `Level::` constants. If all is clear, let's get this merged.

tests/LevelTest.php
ArturMoczulski: @olafnorge this tests that an exception is thrown if a static method is called on `Level` that does not match any of the logging levels. Mostly this doesn't matter, as static methods are now deprecated.

olafnorge: Ahh, thanks for clarifying.

PR #203, opened 2017-06-13
fix tests by getting rid of isolation

rokob:
We do not need process isolation to get rid of the log spew and test interdependencies, we just need to change the tests slightly. This also dramatically speeds up tests.


coryvirok:
Niiiiiiice

Issue #200, opened 2017-06-09
Add helper methods for Level


coryvirok:
It'd be really useful to have some helper methods like:

```php
Rollbar::debug(...)
Rollbar::info(...)
Rollbar::warn(...)
Rollbar::warning(...)
# etc...
```

PR #210, opened 2017-06-26
GitHub Issue #200: Add helper methods for Level

ArturMoczulski:
Static helper methods added to Rollbar class as requested in #200


rokob:
LGTM

Issue #206, opened 2017-06-19
Bug in DataBuilder reversing frames


rokob:
We used to actually reverse the frames in 0.18.2 and prior: https://github.com/rollbar/rollbar-php/blob/fafe13f8beb669d6bba0137f3703fa808d50cbb7/src/rollbar.php#L770

at some point a bug was introduced where we call reverse but do not store the result:

https://github.com/rollbar/rollbar-php/blob/master/src/DataBuilder.php#L470

`array_reverse` does not mutate the argument

PR #211, opened 2017-06-26
GitHub Issue #206: Bug in DataBuilder reversing frames

ArturMoczulski:
_rollbar expects most recent call to be last, not first_ - this was lost somewhere in development of 1.0.0. This PR brings back the expected order of frames as requested in #206

### DIFF
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -450,7 +450,6 @@ class DataBuilder implements DataBuilderInterface
 
             $frames[] = $frame;
         }
-        array_reverse($frames);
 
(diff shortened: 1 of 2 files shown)


ArturMoczulski:
Ready for review

rokob:
LGTM

brianr:
👍  We will need to note this in the changelog; for anyone upgrading from 1.x, this will cause their events to be fingerprinted differently by Rollbar.

rokob:
Good point, it will restore the fingerprints from <= 0.18 but be different than >= 1.0.0 until this is released

Issue #171, opened 2017-05-17
why error_sample_rates only use for errors ?


chrisLeeTW:
There are no way to set sampling rate for exceptions, only for errors.

ArturMoczulski:
This is ready for review in https://github.com/rollbar/rollbar-php/pull/215

PR #215, opened 2017-06-29
GitHub Issue #171: why error_sample_rates only use for errors

ArturMoczulski:
This add `exception_samples_rates` config which allows to set sampling rates for exceptions based on exception class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,13 @@ Default: `'production'`
 Default: empty array, meaning all errors are reported.
 </dd>
 
+<dt>exception_sample_rates
+</dt>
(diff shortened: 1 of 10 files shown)


ArturMoczulski:
This is ready for review

README.md
```diff
@@ -404,6 +404,13 @@ Default: `'production'`
 Default: empty array, meaning all errors are reported.
 </dd>
 
+<dt>exception_sample_rates
+</dt>
+<dd>Associative array mapping exception classes to sample rates. Sample rates are ratio out of 1, e.g. 0 is "never report", 1 is "always report", and 0.1 is "report 10% of the time". Sampling is done on a per-exception basis. It also respects class inheritance meaning if Exception is at 1.0 then ExceptionSublcass is also at 1.0, unless explicitly configured otherwise. If ExceptionSubclass is set to 0.5, but Exception is at 1.0 then Exception and all its' subclasses run at 1.0, except for ExceptionSubclass and its' subclasses which run at 0.5. Names of exception classes should NOT be followed with additional `\` for global namespace, i.e. Rollbar\SampleException and NOT \Rollbar\SampleException.
```
rokob: "should NOT be follow with additional" -> "should NOT be prefixed with additional"

src/Config.php
```diff
@@ -417,38 +423,111 @@ public function checkIgnored($payload, $accessToken, $toLog, $isUncaught)
...
+     * exception sampling.
+     *
+     * @param \Exception $toLog
+     *
+     * @return float
+     */
+    public function exceptionSampleRate(\Exception $toLog)
+    {
+        $sampleRate = 1.0;
+
```
rokob: if `$this->exception_sample_rates` is empty which will be often because that is the default, we should not go through the work of gathering the class hierarchy for this exception, we should just return 1.0 early.

Issue #180, opened 2017-05-26
config.scrub_fields is applied to config.person?


drjayvee:
We recently switched from 0.x to 1.1.1 and noticed that new issues have "******" as their person.username.

Consider this config:
```
'scrub_fields' => ['username', 'password'],
'person' => ['id' => 123, 'username' => 'One Two Three']
```
I suspect that `config.scrub_fields` is applied to `config.person`, causing `person.username` to be scrubbed when an error is logged.

I would propose that `person` data be ignored by the scrubber.

rokob:
There is a bit of tradeoff here so it isn't exactly obvious what one should do. We want to scrub the entire payload based on the scrub_fields configuration, special casing the person data will quickly lead to the opposite issue where people will be surprised that the password that they didn't strip out of their user object is not being scrubbed.

One workaround is to use a custom transform to add the person data to the payload after scrubbing has been done, there are examples in the README about how to add these transform functions.

Additionally, it is a bit odd to scrub username generally, but want it in the one case where it probably shows up which is in the person field. We could add an option for a configurable whitelist of keys to ignore scrubbing if we are nested under that key, but this opens other doors.

I think the easiest fix might just be to have an extra configuration option to ignoring scrubbing for the person data which defaults to false so that if that is what you want you can turn it on, but which most people will not want to do.

drjayvee:
Thanks for your detailed reply.

I agree that it's a tradeoff. Maybe the best way forward is to keep the functionality as is but document this? The documentation about `scrub_fields` should probably mention that it's best not to include `username` (or any other `person` keys).

As for my particular problem: I'll go for an easy workaround: using something other than `username` in the login form.

ArturMoczulski:
This is ready for review in https://github.com/rollbar/rollbar-php/pull/216

PR #216, opened 2017-06-29
GitHub Issue #180: config.scrub_fields is applied to config.person

ArturMoczulski:
This adds `scrub_whitelist` config which allows to create exceptions to scrub fields based on an associative array dot notation.

### DIFF
--- a/src/Scrubber.php
+++ b/src/Scrubber.php
@@ -26,12 +26,12 @@ class Scrubber implements ScrubberInterface
         }
         $this->scrubFields = self::$defaults->scrubFields($fromConfig);
     }
-    
+
     public function getScrubFields()
     {
         return $this->scrubFields;
     }
-    
+
     protected function setWhitelist($config)
     {
         $fromConfig = $this->tryGet($config, 'scrubWhitelist');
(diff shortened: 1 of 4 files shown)


ArturMoczulski:
This is ready for review.

src/Scrubber.php
```diff
@@ -24,20 +26,35 @@ protected function setScrubFields($config)
         }
         $this->scrubFields = self::$defaults->scrubFields($fromConfig);
     }
-
+
```
rokob: Get rid of this whitespace

src/Scrubber.php
```diff
@@ -70,12 +87,31 @@ protected function scrubArray(&$arr, $replacement = '*')
...
+            &$scrubberFn,
+            $scrubber,
+            &$path
+) {
+        
+            $parent = $path;
+            $current = !$path ? $key : $path . '.' . $key;
+            
+            if (in_array($current, $scrubber->getWhitelist())) {
+                return;
```
rokob: If `person` is in scrub_fields and `data.person.username` is in the whitelist, then I am pretty sure we would end up with `data.person === '********'` which I think is the correct behaviour that we want, but is not necessarily obvious. This only works at whitelisting things that would otherwise be scrubbed based on the exact key and not a earlier path component. i.e elements of the whitelist `data.a.b.c.d` will only be whitelisted if a, b, and c are not scrub_fields.

rokob:
LGTM

PR #217, opened 2017-06-29
Performance fixes

rokob:


src/Senders/CurlSender.php
```diff
@@ -60,6 +63,54 @@ public function send($scrubbedPayload, $accessToken)
...
+            if (count($this->inflightRequests) <= $max) {
+                break;
+            }
+            curl_multi_select($this->multiHandle); // or do: usleep(10000);
+        }
+    }
+
+    private function maybeSendMoreBatchRequests($accessToken)
+    {
+        $max = $this->maxBatchRequests-count($this->inflightRequests);
```
coryvirok: Style: missing spaces around `-`

src/Senders/CurlSender.php
```diff
@@ -84,4 +135,36 @@ public function setCurlOptions($handle, $scrubbedPayload, $accessToken)
             }
         }
     }
+
+    private function checkForCompletedRequests($accessToken)
+    {
+        do {
+            $curlResponse = curl_multi_exec($this->multiHandle, $active);
+        } while ($curlResponse == CURLM_CALL_MULTI_PERFORM);
+        while ($active && $curlResponse == CURLM_OK) {
```
coryvirok: It's a bit confusing to read this code with 3 while loops. It might be a bit more concise if this was all put into the do/while loop with a switch statement to handle the different responses.

coryvirok: Actually, reading this a bit closer it seems like this should be broken up into different functions.

coryvirok:
Looking good! Let's disregard the CodeClimate checks. I removed the webhook so it shouldn't show up anymore.

coryvirok:
When can we get these changes merged and released?

rokob:
This is ready to be merged, the big change in functionality related to batching is off by default, so if it has bugs (which it may), it is easy to turn off and fix.

rokob:
This is ready to be merged, the big change in functionality related to batching is off by default, so if it has bugs (which it may), it is easy to turn off and fix.

PR #223, opened 2017-07-14
configuration option to disallow calling exec to get git information

### LINKED ISSUE #220: Sorry but calling exec() on production????
Can we have a configurable way of getting rid of this optionally if you want, the sha1 can be passed via env var but way different not assuming you have git installed even on deploy

``` php
private static function getGitBranch()
    {
        try {
            exec('git rev-parse --abbrev-ref HEAD 2> /dev/null', $output);
            if ($output) {
                return $output[0];
            }
            return null;
        } catch (\Exception $e) {
            return null;
        }
    }
```

rokob:
Fixes #220 

Added the `allow_exec` configuration option which defaults to `true` to conform to the prior behaviour. If you set it to false, you can still pass the branch via the `serverBranch` or `branch` configuration option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,16 @@ All of the following options can be passed as keys in the `$config` array.
 Default: `/var/www`
 </dd>
 
+<dt>allow_exec
+</dt>
+<dd>If the branch option is not set, we will attempt to call out to git to discover the branch name
+via the php `exec` function call. If you do not want to allow `exec` to be called, and therefore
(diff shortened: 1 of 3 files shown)


cordoval:
This is great, thanks, usually the branch on a kubernetes deploy because it is production most of the time is master if not all, and the sha1 is already being passed as ASSETS_PARAMETER, so this helps a lot! thanks.

PR #228, opened 2017-07-15
Github Issue 226: Update rollbar.js code snippet

ArturMoczulski:
This PR introduces automatic updates for rollbar.snippet.js through composer and rollbar/rollbar.js repository.


ArturMoczulski:
Ready for review

rokob:
I don't think we want to always be pulling down master from rollbar.js, I think we want to have stable versions for each rollbar-php release

ArturMoczulski:
@rokob Okay, I locked master at rollbar.js 2.1.0 . This is ready to merge if it looks good to you.

cordoval:
can we make this somehow suggested dep? if we don't use this why would we require it?
I am plugging frontend with frontend stuff and backend with backend stuff separately, why mixing them? Or am i missing something, sorry.

ArturMoczulski:
@cordoval us making a good point imho. unfortunately, if we move rollbar.js to suggestions we can't enforce the version at all other than just provide written instructions as of to what version is the best. I think it would be best to keep rollbar.js as a suggestion and let it be required separately in the appropriate version (with a recommendation in composer suggest description)

ArturMoczulski:
@cordoval we are going to go ahead with rollbar.js as a requirement, but might change it to being a suggested package at a later date.

PR #230, opened 2017-07-19
GitHub Issue 229: Make fluentd a suggest requirement rather than mandatory

ArturMoczulski:
This PR makes fluent/logger a suggested package and skips FluentTest if it's not installed.

Ready for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,8 @@ $config = array(
 ?>
 ```
 
+Also you will have to install a suggested package `fluent/logger`.
+
 
 ### Configuration reference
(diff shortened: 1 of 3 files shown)


cordoval:
lgtm 

btw, is fluentd also behaving like the agent? async i mean?

ArturMoczulski:
PHP notifier passes messages to a local `fluentd` through a socket. I'm not an expert on Fluent but from what https://github.com/fluent/fluentd says it does seem to behave as an async agent.

Issue #227, opened 2017-07-14
Add curl_error() to check for log submission failures


and-megan:
developing on Windows using the WAMP stack ("WAMP SERVER 3.0.0"). It turns out curl was unable to connect to Rollbar securely because properly certificates weren't set up properly. Unfamiliar with how certificates work but found several online answers detailing how this issue, which lots of people with WAMP have encountered, can be fixed.

The Stack Overflow answer that helped: https://stackoverflow.com/a/21114601

suggestion: curl_error() is used to check for this issue in the event of a log failing to submit to Rollbar, or a comment is made in the docs. It would certainly make the process of integrating Rollbar easier for new users.

ArturMoczulski:
@and-megan thanks for feedback on this. Do you happen to have any logs produced as a result of this?

cordoval:
Monolog does something similar:

```
<?php

/*
 * This file is part of the Monolog package.
 *
 * (c) Jordi Boggiano <j.boggiano@seld.be>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Monolog\Handler\Curl;

class Util
{
    private static $retriableErrorCodes = array(
        CURLE_COULDNT_RESOLVE_HOST,
        CURLE_COULDNT_CONNECT,
        CURLE_HTTP_NOT_FOUND,
        CURLE_READ_ERROR,
        CURLE_OPERATION_TIMEOUTED,
        CURLE_HTTP_POST_ERROR,
        CURLE_SSL_CONNECT_ERROR,
    );

    /**
     * Executes a CURL request with optional retries and exception on failure
     *
     * @param  resource          $ch curl handler
     * @throws \RuntimeException
     */
    public static function execute($ch, $retries = 5, $closeAfterDone = true)
    {
        while ($retries--) {
            if (curl_exec($ch) === false) {
                $curlErrno = curl_errno($ch);

                if (false === in_array($curlErrno, self::$retriableErrorCodes, true) || !$retries) {
                    $curlError = curl_error($ch);

                    if ($closeAfterDone) {
                        curl_close($ch);
                    }

                    throw new \RuntimeException(sprintf('Curl error (code %s): %s', $curlErrno, $curlError));
                }

                continue;
            }

            if ($closeAfterDone) {
                curl_close($ch);
            }
            break;
        }
    }
}
```

Maybe it is just plugging this into the CurlSender?

PR #231, opened 2017-07-19
GitHub Issue #227: Add curl_error() to check for log submission failures

ArturMoczulski:
This PR injects result of curl_error() into the Response object in case of curl request problems.

Ready for review.


ArturMoczulski:
This is ready for review in PR https://github.com/rollbar/rollbar-php/pull/231

The result of curl_error() is injected into Response object under info property and available through `getInfo()`. @and-megan let me know if this makes diagnosing your issues easier.

PR #239, opened 2017-07-28
all of the senders should return a response

### LINKED ISSUE #232: Call to a member function getUuid() on a non-object
When we try to upgrade from version 0.18.2 to 1.2.0 (with the agent enabled), we get the error:

`FATAL: Call to a member function getUuid() on a non-object in /vendor/rollbar/rollbar/src/Rollbar.php:226`

When we disable the agent, everything works fine. Any ideas?

rokob:
Fixes #232. The top level notifier assumes that calls to log which eventually call send on a sender return a response object. This is true for all of the senders except the agent. This fixes that.

### DIFF
--- a/src/Senders/AgentSender.php
+++ b/src/Senders/AgentSender.php
@@ -2,6 +2,7 @@
 
 namespace Rollbar\Senders;
 
+use Rollbar\Response;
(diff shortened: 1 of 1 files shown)


ArturMoczulski:
LGTM

Issue #236, opened 2017-07-25
Error Duplication


bretto36:
Hi All

Having some issues with this, everytime an Exception is thrown i get 2 items appear in rollbar. See below. How do i make it so only 1 error is thrown

#3632 E_ERROR: Uncaught Exception:: Uncaught Exception: Call to undefined method: 
#3645 Exception: Call to undefined method: 

\Rollbar\Rollbar::init(array(
                'access_token' => 'a3c1d64d84574031ae664a84622c0608',
                'environment' => (isset($_SERVER['environment']) && 'prod' == $_SERVER['environment'] ? 'production' : 'development'),
                'error_sample_rates' => array(
                    E_ERROR => 1,
                    E_WARNING => 0,
                    E_PARSE => 1,
                    E_NOTICE => 0.001,
                    E_CORE_ERROR => 1,
                    E_CORE_WARNING => 0,
                    E_COMPILE_ERROR => 1,
                    E_COMPILE_WARNING => 0,
                    E_USER_ERROR => 1,
                    E_USER_WARNING => 0,
                    E_USER_NOTICE => 0,
                    E_STRICT => 0,
                    E_RECOVERABLE_ERROR => 1,
                    E_DEPRECATED => 0,
                    E_USER_DEPRECATED => 0,
                )
            ));

ArturMoczulski:
Ready for review in PR https://github.com/rollbar/rollbar-php/pull/240

PR #240, opened 2017-07-30
Github Issue #236: Error Duplication

ArturMoczulski:
This fixes uncaught exceptions being reported twice as per issue #236 

Ready for review

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -167,8 +167,8 @@ class Rollbar
             !is_null($last_error) &&
             in_array($last_error['type'], self::$fatalErrors, true) &&
             // don't log uncaught exceptions as they were handled by exceptionHandler()
-            isset($last_error['message']) &&
-            strpos($last_error['message'], 'Uncaught exception') !== 0;
+            !(isset($last_error['message']) &&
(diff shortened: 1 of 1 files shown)


src/Rollbar.php
```diff
@@ -160,6 +160,16 @@ public static function fatalHandler()
...
     }
+    
+    protected static function shouldLogFatal($last_error)
+    {
+        return
+            !is_null($last_error) &&
+            in_array($last_error['type'], self::$fatalErrors, true) &&
+            // don't log uncaught exceptions as they were handled by exceptionHandler()
+            isset($last_error['message']) &&
+            strpos($last_error['message'], 'Uncaught exception') !== 0;
```
rokob: The `!== 0` should be `!== false` because strpos can return 0 if that is the start of the string, but returns false exactly to indicate not found.

rokob: Oh I see what you are doing, the logic is a bit off for having two ands because you want to log the fatal if message is not set also

ArturMoczulski: Yeah. @rokob thanks for keeping your eyes open though

Issue #241, opened 2017-07-31
Project size


ikovalyov:
Hi.

```
root@997e96a60309:/opt/webapp/rollbar-php# composer update --no-dev
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Loading composer repositories with package information
Updating dependencies
Package operations: 3 installs, 0 updates, 0 removals
  - Installing rollbar/rollbar-js (v2.1.0): Cloning v2.1.0
  - Installing psr/log (1.0.2): Loading from cache
  - Installing fluent/logger (v1.0.1): Loading from cache
Writing lock file
Generating autoload files
root@997e96a60309:/opt/webapp/rollbar-php#ls -l
total 154
-rwxr-xr-x 1 root root 12665 Jul 31 13:59 CHANGELOG.md
-rwxr-xr-x 1 root root  1070 Jul 31 13:59 LICENSE
-rwxr-xr-x 1 root root 12476 Jul 31 13:59 POST_FORMAT.md
-rwxr-xr-x 1 root root 19403 Jul 31 13:59 README.md
-rwxr-xr-x 1 root root  1063 Jul 31 13:59 THANKS.md
-rwxr-xr-x 1 root root   314 Jul 31 13:59 TODO.md
-rwxr-xr-x 1 root root  1836 Jul 31 13:59 composer.json
-rwxr-xr-x 1 root root 89975 Jul 31 14:02 composer.lock
-rwxr-xr-x 1 root root   706 Jul 31 13:59 phpunit.xml
drwxr-xr-x 2 root root     0 Jul 31 13:59 src
drwxr-xr-x 2 root root     0 Jul 31 13:59 tests
drwxr-xr-x 2 root root     0 Jul 31 14:02 vendor
root@997e96a60309:/opt/webapp/rollbar-php# du -sh .
31M     .
```
Don't you think 31 MB for a project wichi sends error logs to 3rd-arty service is a lot?
Thanks.

rokob:
This is due to the inclusion of rollbar.js in vendor, I'm going to see how we can fix that.

Issue #226, opened 2017-07-14
Update rollbar.js code snippet


Guite:
Currently https://github.com/rollbar/rollbar-php/blob/master/data/rollbar.snippet.js includes version 1.9.4 of rollbar.js. The most recent version is 2.1.0 though as shown in https://github.com/rollbar/rollbar.js/releases

Maybe think about a way to specify the desired version or semi-automate the update...

ArturMoczulski:
This is a good place to figure out the process for automating rollbar.js updates.

ytn3rd:
I don't like the idea of the JS code automatically updating because the php code detected a new version, but rather when you deploy rollbar-php it will have a set version of the rollbar.js version. At least then when you test it on a staging server and it works, you know its not going to suddenly break tomorrow because there was a breaking change in the JS that was updated.

Is it possible to have the CI drag in latest released rollbar.js for releases of rollbar-php?

ArturMoczulski:
Ready for review in https://github.com/rollbar/rollbar-php/pull/228

ytn3rd:
Learning more about composer every day, that is heaps better than the silly things I was thinking of!

ArturMoczulski:
Just to clarify, you would like to be able to specify what version of rollbar.js you want to pull in rather than have the PHP SDK pull in the latest by default, is that correct?

ytn3rd:
@ArturMoczulski , not lock to a specific version. My idea was to file_get_contents on a URL to drag latest copy down and save it locally. I didn't like the idea of testing my code and having it work and then having the system auto-update and it breaks. 

Having the rollbar.js as a dependency in composer is a lot better. It means that when I update my packages and test it al, thats where it will be locked at until I update my packages again.

ytn3rd:
Is it possible to set a dependency of roller.js as something like >=1.9.4 or something so it can use anywhere from 1.9.4 to current master, and then if someone wants to stick with 1.9.4 they can add roller.js dependency in their composer manually before rollbar-php so if they want to lock to lower versions they can, if they want master automatically they can.

cordoval:
we should just set some hook to compile this and plug it, no need to stain the composer.json please 👎

rokob:
Including the entire rollbar.js project for just a single small file is untenable. We don't need to auto update the file, just having a script that grabs the snippet when we do a release is good enough.

PR #242, opened 2017-07-31
don't require rollbar.js via composer

rokob:
We don't need the whole rollbar.js repo for just the one snippet file. The only time we need to update that file is during a release. If people want to run a different version of the snippet, then can manually replace the file in their vendor'd repo at data/rollbar.snippet.js. Otherwise, we should just update this snippet when we release new versions of this library. I added a composer script to just pull down the latest rollbar.snippet.js file from github to make that step a bit easier.


rokob:
#241 and #226 are the related issues

PR #247, opened 2017-08-04
Out of the box support for Symfony 2.x and 3.x

sven-communitales:
RollbarHandlerFactory for easy integration into Symfony 2.x and 3.x.

Also updated the README.md with an example.
So no additional Bundles are required to integrate into Symfony.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,39 @@ Rollbar::init(array(
 ?>
 ```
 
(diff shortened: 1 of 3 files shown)


sven-communitales:
Unfortunately the PsrHandler does not support variable replacement.

e.g. `Matched route "{route}".`

PR #255, opened 2017-08-24
Update rollbar.js to the most recent version

nekofar:


rokob:
Can you run that command also to update the stored snippet?

Issue #259, opened 2017-09-01
Remove shiftFunction for exception stack traces


ArturMoczulski:
The `shiftFunction` configuration is causing stack traces for exceptions to inject inaccurate `function` data.

```
array
{
  ...
[7] =>
  class Rollbar\Payload\Frame#31 (9) {
    private $filename =>
    string(72) "/home/ubuntu/workspace/vendor/phpunit/phpunit/src/Framework/TestCase.php"
    private $lineno =>
    int(768)
    private $colno =>
    NULL
    private $method =>
    string(28) "ReflectionMethod::invokeArgs"
    private $code =>
    NULL
    private $context =>
    NULL
    private $args =>
    NULL
    private $kwargs =>
    NULL
    private $utilities =>
    class Rollbar\Utilities#32 (0) {
    }
  }
  [8] =>
  class Rollbar\Payload\Frame#29 (9) {
    private $filename =>
    string(72) "/home/ubuntu/workspace/vendor/phpunit/phpunit/src/Framework/TestCase.php"
    private $lineno =>
    int(909)
    private $colno =>
    NULL
    private $method =>
    string(40) "Rollbar\DataBuilderTest::testFramesOrder"
    private $code =>
    NULL
    private $context =>
    NULL
    private $args =>
    NULL
    private $kwargs =>
    NULL
    private $utilities =>
    class Rollbar\Utilities#30 (0) {
    }
  }
  [9] =>
  class Rollbar\Payload\Frame#27 (9) {
    private $filename =>
    string(48) "/home/ubuntu/workspace/tests/DataBuilderTest.php"
    private $lineno =>
    int(816)
    private $colno =>
    NULL
    private $method =>
    string(6) "<main>"
    private $code =>
    NULL
    private $context =>
    NULL
    private $args =>
    NULL
    private $kwargs =>
    NULL
    private $utilities =>
    class Rollbar\Utilities#28 (0) {
    }
  }
}
```

The last frame (#9) shows `$method` as `<main>` despite the actual method name called being available (overrides it).

Frame #8 shows `Rollbar\DataBuilderTest::testFramesOrder` as `$method` which doesn't match the method actually called in `/home/ubuntu/workspace/vendor/phpunit/phpunit/src/Framework/TestCase.php:909`. Method `Rollbar\DataBuilderTest::testFramesOrder` should actually be listed in frame #9.

[...]

PR #261, opened 2017-09-04
GitHub Issue #259: Remove shiftFunction for exception stack traces

ArturMoczulski:
* remove cutting off backtraces at Rollbar internals - it was hiding essential information in some cases
* remove `shiftFunction` configuration which was causing inaccurate function names in the traces
* extensive testing of exception stack traces and error back traces

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,13 +521,6 @@ Default: `('passwd', 'password', 'secret', 'confirm_password', 'password_confirm
 <dd>Array of fields that you do NOT to be scrubbed even if they match entries in scrub_fields. Entries should be provided in associative array dot notation, i.e. `data.person.username`.
 </dd>
 
-<dt>shift_function
-</dt>
-<dd>Whether to shift function names in stack traces down one frame, so that the function name correctly reflects the context of each frame.
-
-Default: `true`
-</dd>
-
 <dt>timeout
 </dt>
 <dd>Request timeout for posting to rollbar, in seconds.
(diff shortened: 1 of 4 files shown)


rokob:
This seems reasonable to me if it really isn't necessary. I guess I am confused why if we are cutting off vital bits of information then why hasn't everyone been complaining about this? This will also change backtraces for existing errors and will thus change everyone's fingerprinting, so when we release this we will need to note that in the CHANGELOG

ArturMoczulski:
`shiftFunction` caused function names to be moved up in the stack trace which is an inconvenience, but it's not a the end of the world; you can still read through your stack trace. You just have to remember that the filename and line in the frame is actually invoking the function in the frame one up. Because of this I don't think it was a blocker for anybody.

Cutting off through `array_slice` was only introduced in recent releases and only applies to errors and not exceptions. Additionally the only piece of information cut off in there would be the function name of the wrapping function, which is not that big of a deal as well, because the filename and line are still provided. I decided that there is no need to hide information from out stack traces, especially keeping in mind that there will only be 1 or 2 frames from inside `rollbar/rollbar-php`. I'd rather opt-in for full transparency than trying to figure out where to cut off information in this case.

This PR deals with inconveniences of the current stacktraces, rather than blocking problems.

rokob:
I know that we frequently get reports of errors in our code if any rollbar code shows up in the stack trace. This was probably to mitigate the number of erroneous bug reports. But nonetheless this is probably the right change.

Issue #256, opened 2017-08-29
Version 1.3.1 continues to have massive performance issues.


nandgate7400:
I made a comment before I realized the previous issue was marked as closed.  I tried using version 1.3.1 of the library, however my server load goes up by a factor of 20 or more if I just include the init method.  At this point the library is no longer usable in a production environment.  I'm happy to provide any other details if it will help track down the performance issues.

coryvirok:
Thanks @nandgate7400, we will get right on this.

Lidbetter:
Related issue:
https://github.com/rollbar/rollbar-php/issues/209

I have profiled our application with blackfire on 0.18.2 and 1.3.1

On 1.3.1 the error handler callback (`set_error_handler('Rollbar\Rollbar::errorHandler');`) is called ~160 times accounting for ~1 second of CPU time 

On 0.18.2 the error handler callback (`set_error_handler('Rollbar::report_php_error');`) is not called

The php.ini and reporting level was the same in both cases and both libraries were tested on the same machine (one after the other)

I am happy to share our blackfire results (not publically) and have emailed support with a link to this issue.

rokob:
Well that's something, let me see if I can figure that out.

Lidbetter:
I am fairly new to benchmarking with blackfire, I am not 100% sure that it list all calls or just those which have significant impact; if the latter 0.18.2 has no significant impact

rokob:
`set_error_handler` should only ever get called upon the first call to `init`. What is the denominator of that 160 times, i.e. is this a trace with 160 requests or something and `init` gets called once per request?

Lidbetter:
Sorry I meant the callback method name passed into `set_error_handler` is called ~160 times on 1.3.1 vs 0 on 0.18.2 (or as I said it maybe that it has no significantly measurable performance impact)

Lidbetter:
The exact number is probably irrelevant, our error reporting level was the same when testing each version

Lidbetter:
Here is a benchmark of the two libraries, specifically how they handle undefined indexes.

https://github.com/Lidbetter/rollbar-benchmark

A simple fix maybe to use the current error level as the second arg when setting up the error handler (or have it definable in the config)

rokob:
I just merged #258 into master with some clear performance gains by not doing work that we are going to end up throwing away due to a level setting. These are my results from your benchmark:

```
Rollbar Version: 0.18.2
undefinedIndex: 0.050498008728027
[...]

PR #262, opened 2017-09-05
Moar perf

rokob:
Remove some functions because apparently function call overhead is somewhat high in PHP


rokob:
For #256

ArturMoczulski:
Getting rid of `tryGet` and `coalesce` is a great idea. LGTM.

PR #276, opened 2017-10-27
call previous exception handler

flachesis:
restore_exception_handler has no effect in exception handler,
so call previous exception handler immediately.

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -8,6 +8,7 @@ class Rollbar
      * @var RollbarLogger
      */
     private static $logger = null;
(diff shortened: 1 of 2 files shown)


rokob:
After reading the docs, this change seems right to me. If there was another one set when we set ours then we restore it and call it rather than trying to get to it via rethrowing. Otherwise restore wouldn't do anything anyway so we don't have to call it and we can just rethrow. @ArturMoczulski any thoughts?

ArturMoczulski:
Yeah, this looks good to me. Merging this now.

Issue #285, opened 2017-11-04
Dynamically Add global Context


ArturMoczulski:
This is a request originating from https://github.com/rollbar/rollbar-php-laravel, but it applies to https://github.com/rollbar/rollbar-php itself.

The original issue is here https://github.com/rollbar/rollbar-php-laravel/issues/28.

ArturMoczulski:
Closing. This is actually supported with `custom` configuration option.

ArturMoczulski:
This is now ready for review here https://github.com/rollbar/rollbar-php/pull/286

PR #286, opened 2017-11-06
GitHub Issue #285: addCustom() for adding dynamic global context

ArturMoczulski:
This PR accommodates feature request https://github.com/rollbar/rollbar-php-laravel/issues/28

Ready for review and merge.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -240,11 +240,14 @@ class Config
         }
     }
 
-    private function setCustom($config)
+    public function setCustom($config)
     {
-        if (array_key_exists('custom', $config)) {
-            $this->custom = $config['custom'];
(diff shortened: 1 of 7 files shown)


michaelcurry:
There may also be a use case where someone wants to remove a kay from Custom without having to set everything in custom.

michaelcurry:
Having full visibility into the Custom data array would be very benifitial.  In the current use case, removing a key would be very difficult.  There also may be acase where we want to get information from custom before overwriting the key.  The following Function I could se being very useful.

`public function addCustom($key, $data)`


`public function removeCustom($key)`


`public function getCustom($key = null)`
if key is null, return the entire custom array

Issue #297, opened 2017-12-04
Add config option to enable/disable


and-megan:
Can we add a configuration option to easily turn on and off the integration? The idea is to be able to disable Rollbar without having to change or comment out code. Similar to rollbar.js's: 
https://github.com/rollbar/rollbar.js#disable-reporting-to-rollbarcom

nicogranelli:
+1

ArturMoczulski:
This is ready for review in PR https://github.com/rollbar/rollbar-php/pull/300

PR #300, opened 2017-12-10
GitHub Issue #297: add enable / disable config option

ArturMoczulski:
In response to @and-megan request https://github.com/rollbar/rollbar-php/issues/297 I added `::enable()`, `::disable()`, `::configure()` to Rollbar class. This supports easily enabling / disabling Rollbar at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,13 +314,6 @@ Also you will have to install a suggested package `fluent/logger`.
 All of the following options can be passed as keys in the `$config` array.
 
   <dl>
-
-<dt>enabled
-</dt>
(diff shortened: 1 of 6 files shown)


ArturMoczulski:
Disabling Rollbar at runtime will now be available the same way as it is in `rollbar.js`:

```php
Rollbar::configure(['enabled' => false]);
```
or simpler

```php
Rollbar::disable()
```

The same for enabling:

```php
Rollbar::configure(['enabled' => true);
```
or simpler

```php
Rollbar::enable();
```

ftrrtf:
@ArturMoczulski Sorry if I'm wrong, but I do not see `enabled` flag being used anywhere except tests. Did I miss something?

PR #323, opened 2018-02-28
Update MinBodyStrategy.php

khandieyea:
`frames` doesn't always exist, especially when reporting `Rollbar::info` etc. When the payload data is huge, MinBodyStrategy throws warnings on the assumption of `frames`

### DIFF
--- a/src/Truncation/MinBodyStrategy.php
+++ b/src/Truncation/MinBodyStrategy.php
@@ -34,11 +34,13 @@ class MinBodyStrategy extends AbstractStrategy
         /**
          * Limit trace frames
          */
-        $framesStrategy = new FramesStrategy($this->truncation);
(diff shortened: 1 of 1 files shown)


passcod:
I'll also add that, earlier in the code, we see this:

```php
        $traceData = null;
        
        if (isset($payload['data']['body']['trace'])) {
            $traceData = &$payload['data']['body']['trace'];
        } elseif (isset($payload['data']['body']['trace_chain'])) {
            $traceData = &$payload['data']['body']['trace_chain'];
        }
```

That seems like a bug waiting to happen.

If, for some reason, neither of these branches run, `$traceData` will be `null`, and the rest of processing will explode because it expects it to be an array or whatnot.

ArturMoczulski:
@khandieyea great PR. Just merged it. @passcod good observation. I'll amend that in `master`

Issue #265, opened 2017-09-14
invalid or missing instance


paullaffitte:
I'm upgrading a project from 0.18.2 to 1.3.1 and it's working a few times until Rollbar::log start to return an error "invalid or missing instance", with a 400 staus. Then, it continue to return the same error and doesn't log nothing.
I saw a similar issue (#237), but it doesn't help me to solve this.

If it can be useful, here is a var_dump of my $_SESSION just before I try to log
```
array (size=6)
  'useragent6' => 
    array (size=16)
      'browser' => string 'Chrome' (length=6)
      'browser_maker' => string 'Google Inc' (length=10)
      'comment' => string 'Chrome 61.0' (length=11)
      'crawler' => boolean false
      'device_pointing_method' => string 'mouse' (length=5)
      'device_type' => string 'Desktop' (length=7)
      'majorver' => int 61
      'minorver' => int 0
      'parent' => string 'Chrome 61.0' (length=11)
      'platform' => string 'Linux' (length=5)
      'version' => float 61
      'browser_name_pattern' => string 'Mozilla/5.0 (*Linux*) applewebkit* (*khtml*like*gecko*) Chrome/61.0*Safari/*' (length=76)
      'browser_name_regex' => string '/^Mozilla\/5\.0 \(.*Linux.*\) applewebkit.* \(.*khtml.*like.*gecko.*\) Chrome\/61\.0.*Safari\/.*$/' (length=98)
      'ismobiledevice' => boolean false
      'istablet' => boolean false
      'useragent' => string 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.79 Safari/537.36' (length=104)
  'ip_b' => string '��' (length=4)
  'fingerprint2_b' => string '�H~b�lɬ����' (length=16)
  'user_fingerprint_id' => string '89' (length=2)
  'subscribed' => boolean true
  'userData' => 
    array (size=2)
      '_ip' => string '172.18.0.1' (length=10)
      'data' => 
        array (size=3)
          'Username' => string 'Guess' (length=5)
          'ID' => int 0
          'Password' => int 0
```
Edit:
I had a problem also with the previous version we were using (0.18.2). At least the function report_message was triggering a notice "iconv(): Detected an incomplete multibyte character in input string".
Now I "solved" both issues by clearing the cache of my browser.

rokob:
[...]

PR #324, opened 2018-03-06
Fix for issue 265.

h0gar:
See https://github.com/rollbar/rollbar-php/issues/265

Do not embed frame's arguments if they can't be json encoded.

### DIFF
--- a/src/Payload/Frame.php
+++ b/src/Payload/Frame.php
@@ -96,6 +96,9 @@ class Frame implements \JsonSerializable
 
     public function jsonSerialize()
     {
+        $args = $this->args;
+        if(json_encode($args) === false)
+            $args = [];
         $result = array(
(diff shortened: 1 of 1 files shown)


ArturMoczulski:
@h0gar can you provide us with a dump of arguments that are causing the error? I need to build tests for this

ArturMoczulski:
Thanks @h0gar for this invaluable insight. Your PR helped uncover even bigger issue than just the frames arguments. Having an un-encodable value anywhere in the payload would result in `400` response from the API.

I prepared a PR based on your work that resolves this issue: https://github.com/rollbar/rollbar-php/pull/332

h0gar:
Sorry for not taking the time to provide you with more information and a dump of arguments, and thanks for resolving the bigger issue.

PR #332, opened 2018-03-16
GitHub Issue #265: invalid or missing instance

ArturMoczulski:
This PR addressed https://github.com/rollbar/rollbar-php/issues/265

If any of the data in the payload was not encodable to JSON, `json_encode` would result in `data` key of the whole payload to be `false` triggering a `400` response from the API.

Now `json_encode` will simply replace un-encodable values with `null` and still return a valid JSON for ther rest of the structure.


ArturMoczulski:
This is not supported on PHP <= 5.4

Issue #353, opened 2018-04-26
Only collect person.id by default for person tracking


rivkahstandig3636:
Allow the ability to send person.email and person.username, but by default we should only collect person.id going forward.

@ArturMoczulski  This is a top priority.

sugalvojau:
I think details of this case should be added into documentation: 
https://docs.rollbar.com/docs/person-tracking/

I have been trying to make JavaScript and PHP tracking work today. However, I've been stuck on sending username and email from PHP to Rollbar... JavaScript was working fine. Thought something wrong with the PHP code of mine, checked all over.. 

It seems I just had to add `capture_username` and `capture_email` to the `init` config.

```
Rollbar::init(
    [
        'access_token' => accesstokenofmyrollbarproject,
        'environment' => development,
        'capture_username' => true,
        'capture_email' => true,
        'person_fn' => function () {
            return [
                'id' => 123, // required - value is a string
                'username' => 'username123', // optional - value is a string
                'email' => 'user123@example.com' // optional - value is a string
            ];
        }
    ]
);
```

p.s. I don't know the entire history behind this case (smells like GDPR or collision of data when pushing huge data objects to `Rollbar::init` without transformation/validation/etc...) but I have to say this update looks not right. 
If I add data to the bag then I expect/want it to be sent. If I do not add data to the bag then I do not expect/want it to be sent. Sounds quite simple for me.. What is the point of adding user data to the bag and not sending it?
Anyways, as I said, I may not know the entire history of this case. :)

Thanks rivkahstandig36 for starting this topic! 
And thanks to rokob for giving (the hint for) the solution!

PR #363, opened 2018-05-12
GitHub Issue #353: Only collect person.id by default for person tracking

ArturMoczulski:
PR addressing https://github.com/rollbar/rollbar-php/issues/353

Changes the default behavior of the SDK to not collect person's email and username by default. Adds `capture_email` and `capture_username` config options to override the new default behavior if needed.


ArturMoczulski:
@rokob what do you think? This is the closest I could get to replicating what you did in pyrollbar. There will be a separate PR for `capture_ip`.

rokob:
👍

PR #365, opened 2018-05-12
GitHub Issue #354: Allow IP collection to be easily turned on and off in config

ArturMoczulski:
PR addressing https://github.com/rollbar/rollbar-php/issues/354 and https://github.com/rollbar/rollbar-php/issues/355

Adds `capture_ip` which allows turning off gathering the user IP information. Supports value `anonymize` for IP anonymization.

### DIFF
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -811,21 +811,16 @@ class DataBuilder implements DataBuilderInterface
         $ip = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : null;
         
         if ($this->captureIP === DataBuilder::ANONYMIZE_IP) {
-            
             if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4)) {
-                
                 $parts = explode('.', $ip);
                 $ip = $parts[0] . '.' . $parts[1] . '.' . $parts[2] . '.0/24';
-                
-            } else if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
-                
+            } elseif (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
                 $parts = explode(':', $ip);
-                $ip = 
-                    $parts[0] . ':' . 
-                    $parts[1] . ':' . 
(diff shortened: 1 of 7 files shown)


src/DataBuilder.php
```diff
@@ -798,20 +808,39 @@ protected function getRequestBody()
...
+                $parts = explode('.', $ipAddress);
+                $ipAddress = $parts[0] . '.' . $parts[1] . '.' . $parts[2] . '.0/24';
+            } elseif (filter_var($ipAddress, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+                $parts = explode(':', $ipAddress);
+                $ipAddress =
+                    $parts[0] . ':' .
+                    $parts[1] . ':' .
+                    $parts[2] . ':' .
+                    '0000:0000:0000:0000:0000';
+            }
```
rokob: If we get here with an address which `filter_var` determines is not an IPV4 or IPV6 address then we will end up just returning the original `$ipAddress` which seems reasonable because we don't really know how to anonymize something that is not an IP address.

src/DataBuilder.php
```diff
@@ -790,20 +792,44 @@ protected function getRequestBody()
...
+            $ip = $realIp;
+        }
+        $ip = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : null;
+        
+        if ($this->captureIP === DataBuilder::ANONYMIZE_IP) {
+            
+            if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4)) {
+                
+                $parts = explode('.', $ip);
+                $ip = $parts[0] . '.' . $parts[1] . '.' . $parts[2] . '.0/24';
```
andrefedalto: @ArturMoczulski I just noticed that if you replace the last 8 bits of human-readable ip with "0/24", rollbar won't index it and you won't be able to see other errors logged from the same ip. My suggestion would be dropping the "/24" mask and leaving only the "0" at the end. Please see attached image where the addressed with the mask are not clickable.
image

rokob: Yeah I forgot to mention this before this got merged, we need to change the .0/24 to .0 That’s what I ended up doing for the rest of the SDKs

Issue #374, opened 2018-06-06
Rollbar fails to get git branch if allow_exec false and no default set


emporium-nsalter:
This bug was introduced in the 1.5.2 update with the refactoring of the Defaults class. This code in particular (src/Defaults:129): 

```
public function gitBranch($gitBranch = null, $allowExec = true)
    {
        if ($gitBranch) {
            return $gitBranch;
        }
        if (!isset($this->defaultGitBranch) && $allowExec) {
            $this->defaultGitBranch = self::getGitBranch();
        }
        return $this->defaultGitBranch;
    }
```

causes an E_NOTICE to be thrown by line 137 if `$allowExec` is false, as `defaultGitBranch` is never set. There are many cases where a default git branch makes no sense (production, non-git repo) and where exec should not be called. A simple fix would be to set a default value of empty string or 'none' to stop this happening in future.

rivkahstandig3636:
Thanks for letting us know!

PR #378, opened 2018-06-12
Fixed E_NOTICE: Undefined property: Rollbar\Defaults::$defaultGitBranch

vilius-g:
When `'allow_exec'` is set to `false`, `E_NOTICE` errors are generated with every request due to `$this->defaultGitBranch` variable not being defined because of false condition.

This change implements more robust value caching that is used only when `$allowExec` parameter is set to `true`, which is a less suprising behavior.

### DIFF
--- a/src/Defaults.php
+++ b/src/Defaults.php
@@ -131,10 +131,16 @@ class Defaults
         if ($gitBranch) {
             return $gitBranch;
         }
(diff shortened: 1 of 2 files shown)


vilius-g:
Fixes https://github.com/rollbar/rollbar-php/issues/374

ArturMoczulski:
This looks great. Thank you for PR and tests.

PR #390, opened 2018-07-28
Set fingerprint and title using string instead of callback.

brandonlim-hs:
Configure Rollbar fingerprint and title using string instead of function call.
`Rollbar::configure(['fingerprint' => 'custom-fingerprint']);`
`Rollbar::configure(['title' => 'custom-title']);`

### DIFF
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -282,19 +282,11 @@ class DataBuilder implements DataBuilderInterface
     protected function setFingerprint($config)
     {
         $this->fingerprint = isset($config['fingerprint']) ? $config['fingerprint'] : null;
-        if (!is_null($this->fingerprint) && !is_callable($this->fingerprint)) {
-            $msg = "If set, config['fingerprint'] must be a callable that returns a uuid string";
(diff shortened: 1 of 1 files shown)


ArturMoczulski:
@brandonlim-hs what's the motivation behind this? I believe providing a callable that returns a string literal would have the same effect?

brandonlim-hs:
@ArturMoczulski DataBuilder->getFingerprint() does not call the function stored in $fingerprint.
```
Rollbar::configure([
    'fingerprint' => function () {
        return 'custom-fingerprint';
    }
]);
```
Would result with `API error: Invalid format. data.fingerprint should be string.`

Furthermore, [Rollbar docs](https://docs.rollbar.com/docs/grouping-algorithm#section-php) mentioned the support for custom string fingerprints:
```
Rollbar::log(Level::INFO, "Something went wrong", array('fingerprint' => 'hijkl'));
```
Of which, would be nice to keep the configuration of fingerprints consistent?

ArturMoczulski:
@brandonlim-hs gotcha, merging. Thank you for PR

PR #392, opened 2018-08-03
Bug fix for FluentSender 

gveraa:
Missed definition of $scrubbedPayload on send method on FluentSender


ArturMoczulski:
Something is odd with this PR. Are you sure you based the work on latest master?

Issue #404, opened 2018-08-29
Performance Benchmark for Rollbar:init()


ajtran:
internal: https://app.intercom.io/a/apps/rlisg82q/inbox/inbox/813334/conversations/18056584141

user: We just did some benchmarking in our production environment. With the 0.18.2 rollbar-php library, Rollbar::init() is a sub-millisecond operation. In 1.6.2, it takes an average of ~150 milliseconds, with 99th percentile ~270 milliseconds.

Further measurement and documentation of this as well as performant work if necessary would be a great enhancement to the SDK

ArturMoczulski:
Thanks @joshrai those are great insights. I'm putting the SDK under a magnifying glass to find the culprits.

ArturMoczulski:
@joshrai if you want an immediate boost in performance, setting up `allow_exec` to `false` will boost the SDK init by 30% right away. In the above PR I made that a default in the library.

There is one more performance spike I'm looking at before I merge those changes. Should be coming along in the next couple of days @jessewgibbs

joshrai:
Thanks, @ArturMoczulski -- good to know. I just tried that and am seeing similar results. Just to confirm, doing this means that if we want to continue including our Git branch in log events, we should explicitly set a `branch` entry?

I should also mention, our app is very latency-sensitive. We can't easily afford more than ~5 milliseconds for this call. If there's any other info I can provide to help winnow this down, let me know.

joshrai:
I'm getting more confident that the remaining cost, at least in my env, has something to do with object construction overhead outside the source code itself, and somehow specifically for the `Config` class. I added some benchmarking code from just before calling `new Config()`:

https://github.com/rollbar/rollbar-php/blob/1deec89b870899d12794bb75377ca60359f18723/src/RollbarLogger.php#L22

to the first line of the `Config` constructor itself -- before this:

https://github.com/rollbar/rollbar-php/blob/1deec89b870899d12794bb75377ca60359f18723/src/Config.php#L149

and that interval is tending to account for ~90% of the time taken by `Rollbar::init()`. In fact, as it happens, the remaining time seems to be in the neighborhood of 5 msec, so if we could figure this bit out, it might get us home.

This seems like something that could be specific to the particular HHVM (or PHP) implementation, but I think it'll take some more advanced profiling than what I've been doing to learn more.

ArturMoczulski:
[...]

PR #406, opened 2018-08-31
GitHub Issue 404: Performance Benchmark for Rollbar:init()

ArturMoczulski:
This PR addresses performance problems brought up here https://github.com/rollbar/rollbar-php/issues/404

* Turned off autodetecting Git branches by default and added `autodetect_branch` configuration option resulted in ~30% performance increase. Code branch was a rarely used feature anyway and turning it off by default make the SDK perform faster for most users.


ArturMoczulski:
There is also another performance spike in `<cycle 1>` and Composer autoloading as visible in the attached screenshot. I'm working on understanding those two elements as well and see if there is room for optimization there as well.

screen shot 2018-08-31 at 12 27 44 am

ArturMoczulski:
Needs benchmarking on HHVM

Issue #410, opened 2018-09-26
local_vars_dump with recursive data structures causes infinite recursion


jlesueur:
Unfortunately some of our parameters are recursive. Here is a super simplified example:
```php
$obj = [
  'key1' => 'value',
  'key2' => 'value'
];

$obj['key3'] => &$obj;
```

This causes `Utility::serialize($obj)` to enter an infinite recursion. So if local_vars_dump is on, errors can cause infinite recursion.

I kind of think a simple fix is to implement a maximum depth for the serialize method. Maybe a configurable one, with a default of somewhere between 5 and 10? If that solution sounds ok, I'd be willing to put together a PR for it.

jessewgibbs:
@jlesueur thanks for reporting this issue.

@ArturMoczulski what do you think about the proposal to set a maximum depth for the `serialize()` method?  I imagine we'd want to have it be a configurable setting...

coryvirok:
We also need the serialization code to detect loops. See the other sdks for a reference implementation.

ArturMoczulski:
@jessewgibbs yeah I think that's good

@jlesueur if you want to prepare a PR that would be awesome

@coryvirok I'll check out other SDKs, although I believe introducing default max depth should solve the problem

coryvirok:
It might solve it in terms of the specific bug but the serialization will
be incorrect due to loops. Also, it will have another effect for nested
objects of increasing the payload size.

JohnKendhammer:
Issue has been resolved with PR #414

PR #414, opened 2018-10-12
Allow users to specify a max nesting level

jlesueur:
Solves Issue #410

This PR keeps default behavior unchanged, which is to recurse through all data. By setting a `max_nesting_depth` in the configuration, you can limit number of levels that will be traversed.


jessewgibbs:
@jlesueur thanks for the contribution!

@ArturMoczulski can you review this?

JohnKendhammer:
This looks great. Code will get merged into Master

PR #423, opened 2018-11-28
trace_chain must be array

flachesis:
```
<?php
require __DIR__ . '/vendor/autoload.php';

use Rollbar\Rollbar;

function makeException($initLevel = 0, $level = 150)
{
    if ($initLevel >= $level) {
        return null;
    }
    $initLevel += 1;
    return new Exception("$initLevel exception", 0, makeException($initLevel, $level));
}

$exception = makeException();

$rollbar = Rollbar::init([
    'access_token' => '3051774ae88547b5a418ffa2302e8225',
    'environment' => 'development',
], false, false, false);
Rollbar::report_exception($exception);
```

### DIFF
--- a/src/Truncation/FramesStrategy.php
+++ b/src/Truncation/FramesStrategy.php
@@ -12,16 +12,19 @@ class FramesStrategy extends AbstractStrategy
         $key = false;
         $data = $payload->data();
 
-        if (isset($data['data']['body']['trace_chain']['frames'])) {
+        if (isset($data['data']['body']['trace_chain'])) {
             $key = 'trace_chain';
+            foreach ($data['data']['body']['trace_chain'] as $offset => $value) {
+                $data['data']['body']['trace_chain'][$offset]['frames'] = $this->selectFrames($value['frames']);
+            }
         } elseif (isset($data['data']['body']['trace']['frames'])) {
             $key = 'trace';
+            $data['data']['body'][$key]['frames'] = $this->selectFrames($data['data']['body'][$key]['frames']);
         }
         
         
         
         if ($key) {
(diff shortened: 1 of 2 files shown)


jessewgibbs:
@flachesis thanks for the PR.

@ArturMoczulski can you please review?

ArturMoczulski:
The PR looks good although I'm not exactly sure what's the motivation behind this change. It looks like you might be running into an issue with the current implementation in your particular edge case. Can you describe what it is? From the code, I already see you're trying to use the deprecated `report_exception` method for some reason. Can you give us some more background?

ArturMoczulski:
I'm still confused on how this PR achieves shrinking the payload.

Besides, truncation strategies have been modified in the latest `master` to adhere to the new API limit, so this problem shouldn't occur anymore if you update the SDK. The payload will be truncated to respect the API limit automatically. Can you confirm this is the case?

flachesis:
> I'm still confused on how this PR achieves shrinking the payload.
> 
> Besides, truncation strategies have been modified in the latest `master` to adhere to the new API limit, so this problem shouldn't occur anymore if you update the SDK. The payload will be truncated to respect the API limit automatically. Can you confirm this is the case?

Because we have another problem. We use laravel collection as function parameter, but the collection size is too large. StringsStrategy and FramesStrategy not truncate the args payload. I use custom_truncation to add my custom args truncation, and I reference the FramesStrategy class. I read the https://docs.rollbar.com/reference, I find out the trace_chain in the code and document is not match. this pr is not direct relation my problem.

ArturMoczulski:
I understand now. Thanks for catching this. Great PR. Merged into `master` and will be included in the upcoming release.

PR #436, opened 2019-01-23
fixes #434: case-insensitivity on scrub_fields

### LINKED ISSUE #434: scrub_fields are not case insensitive
According to [the docs](https://docs.rollbar.com/docs/php-configuration-reference), the parameters will be converted to lowercase prior to comparing with `scrub_fields`.  Unfortunately this is not happening as a scrub_field of `my_password` won't scrub `MY_PASSWORD` from the data but a scrub_field of `MY_PASSWORD` will.

hubie:


jessewgibbs:
@hubie thanks for the PR.

@ArturMoczulski can you review?

ArturMoczulski:
@hubie great catch. Thanks for the PR. It looks great. I'm merging into `master` and will include in the upcoming release.

PR #437, opened 2019-02-28
Respect the result of any user-defined error handling

ryanshoover:
## Current Situation

Currently the ErrorHandler will call any user-defined `set_error_handler` after Rollbar has processed all if its needs. However, if Rollbar decides to stop handling the error (if no logger is registered or the error is below the logger's base level) then the user's error handler is never called.

## Proposed Solution

The proposed solution will first call the user-defined `set_error_handler`. If the user's function returns a truthy value then Rollbar stops handling the error. A returned truthy value tells PHP to stop handling the error and is a way for user-defined error handlers to circumvent PHP's default methods.

With this solution, any user-defined error handlers can suppress error messages and Rollbar won't pass them upstream.

## Impetus

This PR comes from efforts to suppress a [known warning](https://core.trac.wordpress.org/ticket/44979) triggered by WordPress in a PHP 7.2 environment. Our custom error handler to keep a harmless warning from flooding client sites is being ignored by Rollbar's error handler.

### DIFF
--- a/src/Handlers/ErrorHandler.php
+++ b/src/Handlers/ErrorHandler.php
@@ -6,14 +6,14 @@ use Rollbar\Payload\Level;
 
 class ErrorHandler extends AbstractHandler
 {
-    
+
     public function register()
     {
         $this->previousHandler = set_error_handler(array($this, 'handle'));
-        
+
         parent::register();
     }
-    
+
     public function handle()
     {
         /**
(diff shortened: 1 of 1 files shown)


jessewgibbs:
@ryanshoover thanks for the PR.

@ArturMoczulski could you review when you have a chance?

ryanshoover:
The failed Travis build seems to be a separate issue from my PR? 
```
The command "composer install --no-interaction" failed and exited with 255 during .
```
I didn't touch composer dependencies

ArturMoczulski:
@ryanshoover great catch! Merging the PR right now.

ArturMoczulski:
Released in `v1.7.5`

PR #440, opened 2019-03-13
Add to config minimumLevel option.

Yozhef:
Hello everybody.

Please add the configuration property of the minimum error level.

My problem is:
I have 2 projects in one, I  have to log errors from `Error` and in another from `Info`. Therefore, I need the option to configure the level of the `class Config -> $options -> in my yml` required error level.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -56,7 +56,8 @@ class Config
         'include_raw_request_body',
         'local_vars_dump',
         'max_nesting_depth',
-        'max_items'
+        'max_items',
+        'minimum_level'
     );
     
     private $accessToken;
--- a/src/Config.php
+++ b/src/Config.php
@@ -57,7 +57,7 @@ class Config
         'local_vars_dump',
         'max_nesting_depth',
         'max_items',
-        'minimum_level'
+        'monolog_minimum_level'
     );
     
     private $accessToken;
@@ -89,7 +89,7 @@ class Config
      * @var FilterInterface
      */
     private $filter;
-    private $minimumLevel;
+    private $monologMinimumLevel;
     /**
      * @var ResponseHandlerInterface
      */
@@ -210,7 +210,7 @@ class Config
         $this->setAccessToken($config);
         $this->setDataBuilder($config);
         $this->setTransformer($config);
-        $this->setMinimumLevel($config);
+        $this->setMonologMinimumLevel($config);
         $this->setReportSuppressed($config);
         $this->setFilters($config);
         $this->setSender($config);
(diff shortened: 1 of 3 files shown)


brianr:
Thanks @yozhef!

@ArturMoczulski is there a way to do this already? If not, this seems like a reasonable request to me.

ArturMoczulski:
@Yozhef thanks for submitting the PR

However, I actually can't figure out what is the feature you need here. The PR only contain adding the `minimumLevel` config option, but it is not actually used.

From what you described though, it seems you want to have conditional error levels depending on the environment or the config file loaded. Did I get that right? I'm not sure how a feature called `minimumLevel` would facilitate that.

Can you guide here what you are actually after?

Yozhef:
@ArturMoczulski I need this parameter on other project Rollbar company [rollbar-php-symfony-bundle](https://github.com/rollbar/rollbar-php-symfony-bundle). On this project error level is harcode on createRollbarHandler function in RollbarHandlerFactory class. I need configure minimum error level on my yaml config.

ArturMoczulski:
Ok. Now I see. Thanks for the explanation. I'll leave some feedback on the PR

src/Config.php
```diff
@@ -56,7 +56,8 @@ class Config
         'include_raw_request_body',
         'local_vars_dump',
         'max_nesting_depth',
-        'max_items'
+        'max_items',
+        'minimum_level'
```
ArturMoczulski: I believe this option would only be used with the Monolog handler. I suggest changing this config option to `monolog_minimum_level`. What do you think?

src/Defaults.php
```diff
@@ -358,4 +363,5 @@ public function maxItems($value = null)
     private $scrubFields;
     private $customTruncation = null;
     private $maxItems = 10;
+    private $minimumLevel = null;
```
ArturMoczulski: We also should probably default this to `\Monolog\Logger::error`

src/Config.php
Yozhef: I think is better because this params is found and maybe people use this. You have a good test how it's work. And very long name `monolog_minimum_level`.

src/Defaults.php
Yozhef: Yes it's good =)

src/Config.php
ArturMoczulski: My concern is with the possible confusion of what this config option might be. This is something that will be only used with Monolog handlers. If we don't make that crystal clear a lot of users might mistakenly assume the config option will apply even if they don't use Monolog. Does it make sense?

Yozhef: @ArturMoczulski I thought and still reworked on your offer.  =)

ArturMoczulski:
Thanks for those changes @Yozhef . Actually, I made a mistake. As you could see we already had the `minimumLevel` property, it just got deprecated as a config option. We just need to bring it back and then use it in the Symfony bundle.

I continue the work on this in https://github.com/rollbar/rollbar-php/pull/442 and will be merging in a moment. Let's continue our conversation there.

ArturMoczulski:

PR #449, opened 2019-04-17
Configurable Exception Raising for rollbar.log

### LINKED ISSUE #448: Re-raise exceptions in test and development environments based on a config option
A commonly requested feature is being able to re-raise exceptions in test and development environments based on a config option. The user wants to use the following code:

```php
try {
  do_something_that_might_raise();
} catch(\Exception $ex) {
  Rollbar::log(Log::ERROR, $ex);
}
```

And have it:

* report to Rollbar and swallow the exception in production,
* report to Rollbar and re-raise the exception in dev or test (to not have false positive tests)

ArturMoczulski:
Fixes #448 

Usage:

```php
$enviroment = "test";

Rollbar::init(array(
  "access_token" => $this->getTestAccessToken(),
  "environment" => $environment,
  "raise_on_error" => $environment == "test"
));

try {
  throw new \Exception();
} catch(\Exception $ex) {
  Rollbar::log(Level::ERROR, $ex);
}
```

Above configuration will cause the exception to be reraised only in the `test` environment

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -149,7 +149,8 @@ class Config
     
     /**
      * @var array An array of environments in which the exception should be
-     * reraised from the logger.
+     * reraised from the logger. See for more details: 
+     * https://github.com/rollbar/rollbar-php/issues/448
      */
     private $reraiseInEnvironments = array();
     
--- a/src/Config.php
+++ b/src/Config.php
@@ -57,7 +57,8 @@ class Config
         'local_vars_dump',
         'max_nesting_depth',
         'max_items',
-        'minimum_level'
+        'minimum_level',
+        'reraise_in_environments'
     );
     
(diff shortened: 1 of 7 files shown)


ArturMoczulski:
This feature should also support passing a parameter to the `log` method to force ignoring reraising per individual calls

waltjones:
LGTM

Side note: in the description example, `Rollbar::log` isn't using the `$ex` that's passed in. If this gets cut/pasted for a doc example, look out for that.

ArturMoczulski:
Thanks for pointing that out @waltjones fixed it

ArturMoczulski:
Once this is released, `raise_on_error` needs to get added to the doc

ArturMoczulski:

PR #450, opened 2019-04-25
Remove commitizen setup; prefer global usage instead

### LINKED ISSUE #477: Scrubbing HTTP_COOKIE
## Issue
Some php hosting sets cookie values via `$_SERVER['HTTP_COOKIE']` which can contain sensitive cookie values which are passed through to rollbar. [DataBuilder::getHeaders()](https://github.com/rollbar/rollbar-php/blob/v2.0.0/src/DataBuilder.php#L786-L804) will pull the cookies into the payload.

Adding `Cookie` to `scrub_fields` configuration mitigated the issue for us, however that got rid of all cookies which may or may not be desirable.

Extra note it'd be nice if `scrub_fields` in php worked similarly to how the javascript sdk works where scrub fields appends to the default list of fields instead of replacing it, with an extra option if you wanted to override the default. see [overwriteScrubFields and scrubFields in the js docs](https://docs.rollbar.com/docs/rollbarjs-configuration-reference#section-reference)

ArturMoczulski:
Fix #477

From the further investigation into commitizen turns out that some steps for global installation of the tool will be required anyway, so there is no need for local dev dependencies at all.

Based on https://github.com/rollbar/rollbar-gem/pull/862


ArturMoczulski:

PR #452, opened 2019-04-25
Increase memory limits for php5.3 in Travis CI

### LINKED ISSUE #451: Travis CI fails for master on PHP 5.3
https://travis-ci.org/rollbar/rollbar-php/jobs/521147979

On PHP 5.3 `composer install` fails due to exceeding memory limits. It seems like a common issue with PHP 5.3. The memory limits need to be increased.

ArturMoczulski:
fix #451

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ matrix:
   include:
     - php: 5.3
       dist: precise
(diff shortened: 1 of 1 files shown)


ArturMoczulski:

PR #456, opened 2019-04-30
Standardized dev options: enabled, transmit, output, verbose

### LINKED ISSUE #453: Standardize enabled, transmit, output and verbose configuration options
There has been much confusion around `enabled` and `verbose` config options available across Rollbar SDKs. It is not clear to the users what each does versus what it is actually doing. We identified four different parameters that should control the behavior of the SDK in regards to it being enabled and how the logging is performed.

The new suggested configuration options authored by @rokob  are:

I propose we have the following settings which are all independent of one another:
- `enabled`
  - If this is false then do absolutely nothing, try to be as close to the scenario where Rollbar did not exist at all in the code.
  - Default: `true`
- `transmit`
  - If this is false then we do everything except make the post request at the end of the pipeline.
  - Default: `true`
- `output`
  - If this is true then we output the payload to standard out or a configured logger right before transmitting.
  - Default: `false`
- `verbose`
  - If this is true then we output messages related to the processing of items that might be useful to someone trying to understand what Rollbar is doing.
  - Default: `false`
    - I would actually prefer this to be a level in increasing level of severity:
      - `trace`
      - `debug`
      - `info`
      - `warn`
      - `error`
      - `none`
    - The SDK would then have internal messages of different log levels and based on what you pick we would output only what you want to see, i.e. trace prints everything, error prints only the worst errors, and none prints nothing.

ArturMoczulski:
Closes #453 

* implements `transmit` (default: `true`) which makes the SDK skip calling `send` and `sendBatch` on `sender` objects
* implements `output` (default: `false`) which makes the SDK dump the payload and `Response` object to the configured `output_logger` (default: `\Monolog\Logger` with `\Monolog\Handler\ErrorLogHandler`)
* implements `verbose` (default: `\Rollbar\Config::VERBOSE_NONE`) which writes all verbose messages to configured `verbose_logger` (default: default: `\Monolog\Logger` with `\Monolog\Handler\ErrorLogHandler`)
  * values accepted by `verbose` config option are `\Psr\Log\LogLevel` values in compliance with https://tools.ietf.org/html/rfc5424 rather than custom config options suggested in the original spec
  * `verbose` supports `\Rollbar\Config::VERBOSE_NONE` which makes the SDK not verbose (except for `output` messages configured by a separate `output` option)
* `output_logger` and `verbose_logger` are independent. `output_logger` is configurable solely with a boolean `output` config option; `verbose_logger` is configurable with `\Psr\Log\LogLevel` verbosity level
* adds verbose messages (when appropriate level of `verbose` is set):
  - `\Rollbar\RollbarLogger::log`
    - `info`: `Attempting to log: [$level] ...` always
    - `notice`: `Rollbar is disabled` when `enabled == false`
    - `error`: `Invalid log level '$level'.` when invalid log level is passed to `RollbarLogger`
    - `info`: `Occurrence ignored` when an occurrence is ignored due to any reason covered by `\Rollbar\Config::internalCheckIgnored`
    - `info`: `Occurrence ignored` when an occurrence is ignored due to any reason covered by `\Rollbar\Config::checkIgnore`
    - `error`: `Occurrence rejected by the SDK: $response` when the SDK rejected the occurrence for any reason resulting in response status `0`
    - `error`: `Occurrence rejected by the API: $response` when the API rejected the occurrence for any reason resulting in HTTP response status code `400+`
    - `info`: `Occurrence successfully logged` always on successful report
  - `\Rollbar\RollbarLogger::send`
    - `warning`: `Maximum number of items per request has been reached. If you want to report more items, please use `max_items` configuration option.` when transmission stopped due to exceeding the `max_items` config option
    - `debug`: `Added payload to the queue (running in `batched` mode).` always when the occurrence is added to the queue (`batched == true`)
  - `\Rollbar\RollbarLogger::flush`
    - `debug`: `Queue flushed` always
  - `\Rollbar\Config::shouldIgnoreError`
    - `debug`: `Ignore (error below allowed error_reporting level)` when `use_error_reporting == true` and the PHP error's level is below configured PHP's `error_reporting` level
    - `debug`: `Ignore due to included_errno level` when PHP error's level is not included in configured `included_errno`
    - `debug`: `Skip due to error sample rating` when PHP error is skipped due to configured `error_sample_rates`
  - `\Rollbar\Config::shouldIgnoreException`
    - `debug`: `Skip exception due to exception sample rating` when an exception is skipped due to configured `exception_sample_rates` 
  - `\Rollbar\Config::checkIgnored`
    - `info`: `Occurrence ignored due to custom checkIgnore logic` when an occurrence is ignored due to `check_ignore` callback evaluating to `true`
    - `error`: `Exception occurred in the custom checkIgnore logic: $exceptionMessage` when an exception is thrown in `check_ignore` callback
    - `debug`: `Occurrence's level is too low` when occurrence's level is below configured `minimum_level`
    - `debug`: `Custom filter result: true/false` always when `filter` is configured
  - `\Rollbar\Config::send`
    - `warning`: `Not transmitting (transmitting disabled in configuration)` when not transmitting due to `transmit == false`
  - `\Rollbar\Config::sendBatch`
    - `warning`: `Not transmitting (transmitting disabled in configuration)` when not transmitting due to `transmit == false`
  - `\Rollbar\Config::internalCheckIgnored`
    - `debug`: `Ignoring (error reporting has been disabled in PHP config)` when PHP's `error_reporting === 0`
    - `debug`: `Occurrence's level is too low` when occurrence's level is below configured `minimum_level`
  - `\Rollbar\Config::handleResponse`
    - `debug`: `Applying custom response handler: get_class($this->responseHandler)` when applying configured `responseHandler`
  - `\Rollbar\Truncation\Truncation::truncate`
    - `debug`: `Applying truncation strategy get_class($strategy)` always when applying a truncation strategy to payload
* tests verbose messages


akornich:
@ArturMoczulski , didn't we want to have a list of "EnvironmentsToTransmitFor" list option instead of the "Transmit" bool option?

waltjones:
@akornich I would prefer boolean wherever possible so the user can define their logic, which may be based on environment, but could be based on other logic. It also makes it more sensible for dynamically changing the value at runtime. This also means the SDK doesn't have to know what environments the app has defined.

See the example in this PR description: https://github.com/rollbar/rollbar-php/pull/449

ArturMoczulski:
The response dump needs to get removed

ArturMoczulski:

PR #459, opened 2019-05-23
Fix incorrect array encoding in config produced by RollbarJsHelper

vilius-g:
When building JS code with RollbarJsHelper and passing configuration
parameters that are arrays, those arrays in the final JS output are
incorrectly encoded as objects.

### DIFF
--- a/src/RollbarJsHelper.php
+++ b/src/RollbarJsHelper.php
@@ -61,7 +61,7 @@ class RollbarJsHelper
      */
     public function configJsTag()
     {
-        return "var _rollbarConfig = " . json_encode($this->config, JSON_FORCE_OBJECT) . ";";
+        return "var _rollbarConfig = " . json_encode((object)$this->config) . ";";
     }
     
     /**
--- a/tests/JsHelperTest.php
+++ b/tests/JsHelperTest.php
@@ -226,19 +226,29 @@ class JsHelperTest extends BaseRollbarTest
         );
     }
     
-    public function testConfigJsTag()
+    /**
+     * @dataProvider configJsTagProvider
+     */
+    public function testConfigJsTag($config, $expectedJson)
     {
-        $config = array(
-            'config1' => 'value 1'
-        );
-        
-        $expectedJson = json_encode($config);
         $expected = "var _rollbarConfig = $expectedJson;";
         
         $helper = new RollbarJsHelper($config);
         $result = $helper->configJsTag();
         
         $this->assertEquals($expected, $result);
+    }
+    
+    public function configJsTagProvider()
+    {
+        return array(
+            array(array(), '{}'),
+            array(array('config1' => 'value 1'), '{"config1":"value 1"}'),
+            array(
+                array('hostBlackList' => array('example.com', 'badhost.com')),
+                '{"hostBlackList":["example.com","badhost.com"]}'
+            ),
+        );
     }
     
     /**


bishopb:
Thanks for taking the time to submit this! I agree; the current code seems buggy. 

Digging a little, looks like [`JSON_FORCE_OBJECT` was introduced back in 2017](https://github.com/rollbar/rollbar-php/commit/b39234b62926bb33fea88478e309dbb9ff6ba3a7#diff-fc2487797c2db4b13b48c6b0b6a1767519ca2865f4fe67c84892d7f8f0108530R59) by @rokob. I _think_ the intent of that change was to ensure the configuration object at the highest level was always an object, but it seems to have neglected the side effect of forcing deep PHP arrays to JSON objects (eg, a configuration of `[ "a" => [ ] ]` produces `{"a":{}}` instead of the desired `{"a":[]}`.

We'll get this in the pipeline for change soon.

src/RollbarJsHelper.php
```diff
@@ -63,7 +63,10 @@ public function addJs($headers = null, $nonce = null, $customJs = "")
      */
     public function configJsTag()
     {
-        return "var _rollbarConfig = " . json_encode($this->config, JSON_FORCE_OBJECT) . ";";
+        return "var _rollbarConfig = " . json_encode(
+            $this->config,
+            $this->config ? null : JSON_FORCE_OBJECT
```
bishopb, line 68: I think the intent of the `JSON_FORCE_OBJECT` was to ensure that the highest level was always an object, even when given as an empty array: eg, `"{}" === (new \Rollbar\RollbarJsConfig([]))->configJsTag()`. Given that, I think it'd be more clear to implement as:

```
return "var _rollbarConfig = " . json_encode((object)$this->config);
```

That ensures the top-most result is a JSON object, while preserving nested PHP arrays.

src/RollbarJsHelper.php
```diff
@@ -63,7 +63,10 @@ public function addJs($headers = null, $nonce = null, $customJs = "")
...
+        ) . ";";
```
bishopb, line 69: I think the intent of the `JSON_FORCE_OBJECT` was to ensure that the highest level was always an object, even when given as an empty array: eg, `"{}" === (new \Rollbar\RollbarJsConfig([]))->configJsTag()`. Given that, I think it'd be more clear to implement as:

```
return "var _rollbarConfig = " . json_encode((object)$this->config);
```

That ensures the top-most result is a JSON object, while preserving nested PHP arrays.

bishopb (changes requested):

rokob:
Yeah I was trying to only force the top-level to be an object. PHP never ceases to amaze me. I see now in the docs that there is a user comment about this behavior. Although the main docs don't mention anything about this being applied to the entire structure:

> JSON_FORCE_OBJECT (int)
> Outputs an object rather than an array when a non-associative array is used. Especially useful when the recipient of the output is expecting an object and the array is empty. Available as of PHP 5.3.0.

danielmorell (changes requested):
@vilius-g sorry about the long delay! This looks great! I think we just need to fix the merge conflicts before getting this in. Can you do that?

danielmorell (approved):
Thank you @vilius-g. I ran the failed tests again and they passed. We sometimes get false negatives when performing the network part of the test suite. 

This looks great! Thank you for the work on this!

PR #468, opened 2019-09-20
support for monolog ^2.0 as well as ^1.0

jonnott:
bonus: allows rollbar-php-laravel to support Laravel 6.0 without having to downgrade monolog explicitly to ^1.0 in your app


ArturMoczulski:
Should be resolved in https://github.com/rollbar/rollbar-php-laravel/releases/tag/v5.0.0

Please let me know if you run into any problems

PR #474, opened 2019-12-03
Fix compatibility with ext-ds

enumag:
Some ext-ds structures like [Map](https://www.php.net/manual/en/class.ds-map.php) can implement Traversable and use objects as keys. This causes `Warning: Illegal offset type` in your library.

### DIFF
--- a/src/Utilities.php
+++ b/src/Utilities.php
@@ -120,7 +120,7 @@ final class Utilities
             
             if ($customKeys !== null && in_array($key, $customKeys)) {
                 $returnVal[$key] = $val;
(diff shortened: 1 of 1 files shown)


bishopb:
Good catch, and thanks for the submission!

Would you mind adding a test to cover this new usage?

See also https://github.com/rollbar/rollbar-php/pull/497 which does some similar, but non-conflicting, work.

enumag:
I've dropped rollbar in favor of sentry, mostly because of inactivity on this issue since I'm using ext-ds extensively so it was a blocker. No longer interested in getting this fixed.

PR #483, opened 2020-04-22
Fix Error when handling invalid reponses from rollbar in php7.4.

lxlang:
Mitigating Array-style access of non-arrays in php 7.4

Prevents ErrorException: Trying to access array offset on value of type null.


lxlang:
Fixes https://github.com/rollbar/rollbar-php/issues/484

ryan-deng-plastiq:
why is this closed, i think it could fix the error?

lxlang:
@ryan-deng-plastiq Was done in https://github.com/rollbar/rollbar-php/pull/485

PR #485, opened 2020-04-29
Fix Error when handling invalid reponses from rollbar in php7.4.

lxlang:
Mitigating Array-style access of non-arrays in php 7.4

Prevents ErrorException: Trying to access array offset on value of type null.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: php
 dist: trusty
 
 php:
+  - 7.4     
+  - 7.3     
   - 7.2
   - 7.1
 
--- a/src/RollbarLogger.php
+++ b/src/RollbarLogger.php
@@ -118,7 +118,7 @@ class RollbarLogger extends AbstractLogger
             $this->verboseLogger()->error('Occurrence rejected by the SDK: ' . $response);
         } elseif ($response->getStatus() >= 400) {
             $info = $response->getInfo();
-            $this->verboseLogger()->error('Occurrence rejected by the API: ' . $info['message']);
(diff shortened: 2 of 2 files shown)


mrunalk:
@icsahn-rollbar, I can update the docs. Thanks @lxlang for adding the support for 7.3 and 7.4. We welcome contributions from the community.

mrunalk:
@lxlang, we had to revert your change due to our travis builds not able to setup for 7.3 and 7.4 properly. But with #486 we have fixed the issue. Can you please test it and see if that solves your issue. Appreciate your feedback!

fgilio:
@mrunalk so this needs to be merged again? I'm having this same issue https://github.com/rollbar/rollbar-php-laravel/issues/99 in with v6.0.0 and PHP 7.4

lxlang:
@mrunalk 

Thanks for your patience. 

This change looks fine for me and I would really appreciate seeing this in a new release. 
Currently we have many problems with our rollbar integration because of this error.

fgilio:
Meanwhile, we resorted to overloading the class in Composer to patch the issue. It's working fine 👍🏼

lxlang:
That's great, but without a new release, I cannot utilize this change in my systems.

fgilio:
> That's great, but without a new release, I cannot utilize this change in my systems.

@lxlang 👋🏼 it's now tagged here https://github.com/rollbar/rollbar-php/releases/tag/v2.1.0

Thanks for the fix

mrunalk:
@lxlang @fgilio thank you for your patience. Yep it's now available as part of 2.1.0 release.

PR #488, opened 2020-05-04
Fixing Travis build for php 7.1 to php 7.4

icsahn-rollbar:
The PR fixes the failure in Travis jobs at various places. It also minor code formatting issues.

Notably:
1. PHP 7.1 requires PHPUnit 5.2.7.
2. `syntaxCheck ` is [deprecated](https://stackoverflow.com/questions/44328114/phpunit-what-does-syntaxcheck-configuration-parameter-stands-for-exactly/44331140#44331140)

### DIFF
--- a/tests/ConfigTest.php
+++ b/tests/ConfigTest.php
@@ -97,32 +97,32 @@ class ConfigTest extends BaseRollbarTest
         $this->assertTrue($config->loggingPayload());
     }
 
-    public function testLoggingPayload()
-    {
-        $logPayloadLoggerMock = $this->getMockBuilder('\Psr\Log\LoggerInterface')->getMock();
-        $logPayloadLoggerMock->expects($this->once())
-                        ->method('debug')
-                        ->with(
-                            $this->matchesRegularExpression(
-                                '/Sending payload with .*:\n\{"data":/'
-                            )
-                        );
-        $senderMock = $this->getMockBuilder('\Rollbar\Senders\SenderInterface')
-                        ->getMock();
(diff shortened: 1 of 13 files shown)


mrunalk:
@icsahn-rollbar, thanks for fixing the travis errors. Can you please update the PR title and description which indicates updates to PHP and other changes we are trying to do here?

tests/ConfigTest.php
```diff
@@ -30,6 +30,7 @@ public function setUp(): void
             null,
             new Utilities
         );
+        parent::setUp();
```
waltjones, line 33: I'm curious about this. I'm noticing some of the test classes call parent in `setUp()`, and some don't. And the implementation in `PHPUnit_Framework_TestCase` is an empty function. Was this breaking before, or is it for style/correctness?

icsahn-rollbar: this is mostly style/correctness piece. Since the base class is overriding the method, it should be calling its parent's setup as well. 

In this particular case, it should change anything functionally. (Since parent's setup is empty)

icsahn-rollbar:
> @icsahn-rollbar, thanks for fixing the travis errors. Can you please update the PR title and description which indicates updates to PHP and other changes we are trying to do here?

Done. Just FYI, this PR will merge with your [branch](https://github.com/rollbar/rollbar-php/tree/php7-4).

mrunalk:
@icsahn-rollbar, LGTM. Thanks for working on fixing the build. Can you update the PR or your branch so that it merges into master then?

PR #494, opened 2020-06-29
Replace Whitelist with Safelist

mrunalk:
This is a BREAKING CHANGE. Consumers currently relying upon `scrub_whitelist` or `setWhitelist` will need to update their code or configuration to `scrub_safelist` and `setSafelist` respectively.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -66,7 +66,7 @@ class Config
         'capture_email',
         'root',
         'scrub_fields',
-        'scrub_whitelist',
+        'scrub_safelist',
         'timeout',
         'transmit',
         'custom_truncation',
--- a/src/Scrubber.php
+++ b/src/Scrubber.php
@@ -6,13 +6,13 @@ class Scrubber implements ScrubberInterface
 {
     protected static $defaults;
     protected $scrubFields;
-    protected $whitelist;
+    protected $safelist;
 
     public function __construct($config)
     {
         self::$defaults = Defaults::get();
         $this->setScrubFields($config);
-        $this->setWhitelist($config);
+        $this->setSafelist($config);
     }
 
     protected function setScrubFields($config)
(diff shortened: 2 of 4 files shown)


src/Config.php
```diff
@@ -50,7 +50,7 @@ class Config
         'capture_email',
         'root',
         'scrub_fields',
-        'scrub_whitelist',
+        'scrub_safelist',
```
waltjones, line 53: I know the assignment here says `$options` is private, but according to the docs the user has access to set this. Will this change affect existing users?

mrunalk: @waltjones, you bring up a good point. I am not totally sure how options are exposed. Let me look into it.

bishopb: Two points:

1. We can _add_ a new key, but we shouldn't _remove_ `scrub_whitelist` at the same time: there should be at least a minor version increment between them.

2. I prefer `allowlist` instead of `safelist`, because the opposite is (IMO) clearer: `denylist` vs `unsafelist`. Is the "safelist" nomenclature a standard in Rollbar?

bishopb:
I have marked this as "Status: 2 - Accepted" because the concept and need is understood; there are still some questions on the implementation, though: see above.

src/Config.php
waltjones: `allowlist` was considered, but the consensus was to standardize on `safelist` (and `blocklist`.)

The CH story for each of the SDKs is to only deprecate, not remove, public facing names until the next major version release.

phpunit.xml
```diff
@@ -17,9 +17,9 @@
     </testsuite>
   </testsuites>
   <filter>
-    <whitelist processUncoveredFilesFromWhitelist="true">
+    <safelist processUncoveredFilesFromSafelist="true">
       <directory suffix=".php">./src</directory>
-    </whitelist>
+    </safelist>
```
bishopb, line 22: This change is not relevant to the implementation. Needs reversion.

akornich (commented):
lgtm

akornich (approved):

waltjones (approved):

PR #497, opened 2020-08-19
Extended Object Normalization

ragboyjr:
- Added better support for serializing common
  types of classes since a lot of classes use
  private fields which make the default object
  normalization basically useless

Signed-off-by: RJ Garcia <ragboyjr@icloud.com>

## Description of the change

> Description here
## Type of change

### Development

- [ ] Lint rules pass locally
- [ ] The code changed/added as part of this pull request has been covered with tests
- [ ] All tests related to the changed code pass in development

### Code review 

### DIFF
--- a/src/Utilities.php
+++ b/src/Utilities.php
@@ -135,31 +135,47 @@ final class Utilities
         $maxDepth = -1,
         $depth = 0
     ) {
-        $serialized = null;
         
         if (self::serializedAlready($obj, $objectHashes)) {
-            $serialized = self::circularReferenceLabel($obj);
-        } else {
-            if ($obj instanceof \Serializable) {
-                self::markSerialized($obj, $objectHashes);
-                $serialized = $obj->serialize();
-            } else {
-                $serialized = array(
-                    'class' => get_class($obj)
-                );
-                
-                if ($obj instanceof \Iterator) {
-                    $serialized['value'] = 'non-serializable';
-                } else {
-                    $serialized['value'] = self::serializeForRollbar(
-                        $obj,
-                        $customKeys,
-                        $objectHashes,
-                        $maxDepth,
-                        $depth+1
-                    );
-                }
+            return self::circularReferenceLabel($obj);
+        }
+
+        if ($obj instanceof \Serializable) {
+            self::markSerialized($obj, $objectHashes);
+            return $obj->serialize();
+        }
+        
+        $serialized = array(
+            'class' => get_class($obj)
+        );
+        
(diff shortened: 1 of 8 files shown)


src/Utilities.php
```diff
@@ -135,31 +135,47 @@ private static function serializeObject(
...
+        );
+        
+        if ($obj instanceof \Iterator) {
+            $serialized['value'] = 'non-serializable';
+            return $serialized;
+        }
+        
+        try {
+            if (method_exists($obj, 'toString')) {
+                $serialized['value'] = $obj->toString();
```
bishopb, line 159: The method `toString` we would _think_ would return a `string`, however that's not guaranteed: someone could do `function toString(): resource { return fopen($this->path, 'r'); }`. Recommend that the output of `$obj->toString()` pass through `serializeForRollbar` to ensure it's serialized. That may require some juggling of the checks.

src/Utilities.php
```diff
@@ -135,31 +135,47 @@ private static function serializeObject(
...
+            } elseif (method_exists($obj, '__toString')) {
+                $serialized['value'] = (string) $obj;
+            } elseif (method_exists($obj, 'toArray')) {
```
bishopb, line 162: Same applies here as it did with the `toString` comment: there's no guarantee that `toArray` actually returns an `array`.

src/Utilities.php
```diff
@@ -135,31 +135,47 @@ private static function serializeObject(
...
+                $serialized['value'] = $obj->toArray();
+            } elseif ($obj instanceof \JsonSerializable) {
+                $serialized['value'] = $obj->jsonSerialize();
             }
+        } catch (\Throwable $e) {
```
bishopb, line 167: `Throwable` is not safe for 5.x, which this version of the Rollbar library supports. We also need a `Exception` block to handle that version.

tests/TestHelpers/ObjectSerialization/JsonSerializable.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Rollbar\TestHelpers\ObjectSerialization;
```
bishopb, line 3: Nit: these may not pass linting, as we have `namespace` beside the `<?php` in other files. However, we will be changing this in the next major iteration of this library in order to support `strict_types`.

bishopb (changes requested):
Thanks for these improvements! They seem like a good value add. There are a few changes necessary, but otherwise we should be good to go.

danielmorell:
I am closing this PR as it has become stale.

PR #499, opened 2020-10-02
Add isUncaught to context array, remove fourth parameter to log in RollbarLogger

jamesaspence:
## Description of the change

PSR-3's `LoggerInterface` has a method, `log`, that accepts three arguments: `$level`, `$message`, and `$context`. The `RollbarLogger` accepts four arguments, the fourth being `$isUncaught`. This PR removes that fourth argument to align with the PSR-3 method signature, and instead adds it to the `$context` array. This ensures that the logger follows the PSR-3 standard.

This would break any direct calls to `RollbarLogger` - I fixed any direct reference to this class here in the src code, but anybody consuming this code outside would need to update their code to match the new parameter structure. If this is a major concern we might be able to find a way to retrieve the fourth argument when it's passed in and throw a warning up while still adding it to context. I'd be happy to make that change if that's desired.

## Type of change

## Related issues
N/A

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/Handlers/ErrorHandler.php
+++ b/src/Handlers/ErrorHandler.php
@@ -58,7 +58,7 @@ class ErrorHandler extends AbstractHandler
                             getDataBuilder()->
                             generateErrorWrapper($errno, $errstr, $errfile, $errline);
 
-        $this->logger()->log(Level::ERROR, $exception, array(), true);
+        $this->logger()->log(Level::ERROR, $exception, array('isUncaught' => true));
 
         return false;
     }
--- a/src/Handlers/ExceptionHandler.php
+++ b/src/Handlers/ExceptionHandler.php
@@ -30,7 +30,7 @@ class ExceptionHandler extends AbstractHandler
         
         $exception = $args[0];
         
-        $this->logger()->log(Level::ERROR, $exception, array(), true);
+        $this->logger()->log(Level::ERROR, $exception, array('isUncaught' => true));
         
         if ($this->previousHandler) {
             restore_exception_handler();
--- a/src/Handlers/FatalHandler.php
+++ b/src/Handlers/FatalHandler.php
@@ -41,7 +41,7 @@ class FatalHandler extends AbstractHandler
                                 getDataBuilder()->
                                 generateErrorWrapper($errno, $errstr, $errfile, $errline);
                                 
-            $this->logger()->log(Level::CRITICAL, $exception, array(), true);
+            $this->logger()->log(Level::CRITICAL, $exception, array('isUncaught' => true));
         }
     }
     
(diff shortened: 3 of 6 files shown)


jamesaspence:
Also - apologies, I mixed up PSR-3 and PSR-7. This brings the logger in line with the PSR-3 spec.

jamesaspence:
Bump? Any interest in this change? I think it's a good change, to bring the package more in line with the logger spec, but I understand if there's not an interest in bringing it in.

bishopb:
Thanks for raising awareness of this difference and proposing a fix! While I agree it's best to keep signatures identical, in this case - because this is a breaking change of unknown scope (zero consumers may be affected, then again maybe a lot are) - I'd like to see a stronger argument made.

As I understand the request, the [PSR-3 signature for `log`](https://www.php-fig.org/psr/psr-3/) accepts exactly three arguments: `$level`, `$message`, and `array $context`, while our derived method `RollbarLogger::log` accepts these three _plus_ a new one: `$isUncaught`. While that violates the letter of the `log` interface, it doesn't violate the _spirit_. Importantly, it also doesn't violate [LSP](https://madewithlove.com/liskov-substitution-principle-explained/): we're keeping the pre-conditions of the `log` method intact, just [widening them](http://serebrov.github.io/html/2016-02-18-oop-solid-l-liskov-substitution-principle.html) to accept our use case.

Absent a strong case that we're losing interoperability because of this, I'm inclined to wait and revisit this at the next major version increment of the library. Thoughts?

jamesaspence:
Totally understand that thought process, and I actually agree with you - I definitely think that introducing this change could cause some legitimate issues for consumers of this class elsewhere.

Would it be reasonable to change my code to leave that fourth argument in, but log a deprecation warning if that fourth argument is presented? We could change the signatures to something like:
```
(bool $isUncaught = null) {
    if (!is_null($isUncaught)) {
        // However we want to handle the notice for deprecation in a logging library
        $this->logDeprecationNotice();
    } else {
        $isUncaught = false;
    }
}
```

If you'd prefer to wait until the next major release instead of something like the above I also totally understand - just happy that there was some movement on the issue!

bishopb:
@jamesaspence I have taken a different approach to solve the underlying issue: removing the `$isUncaught` parameter from the `log()` signature. See https://github.com/rollbar/rollbar-php/pull/518. This change targets release 3.0.0, and will be available in 3.0.0-RC2. Note 3.0 requires PHP 8.

As such, I am closing this pull in favor of the other one. If you'd like to discuss this change further, potentially within the context of the 2.x series, feel free to re-open this PR. Thank you for your contribution!

PR #521, opened 2021-05-25
Annotate function signatures

bishopb:
## Description of the change

Robust API make definitive statements about their acceptable inputs. One way to do that is through documentation (which was being done in many cases), but the better way is through type assertions. In this PR, we add strong types to key classes that have the greatest interaction with consumers.

## Type of change

## Related issues

CH-79868

### Development

- [X] Lint rules pass locally

### Code review 

### DIFF
--- a/src/RollbarLogger.php
+++ b/src/RollbarLogger.php
@@ -1,4 +1,6 @@
-<?php namespace Rollbar;
+<?php declare(strict_types=1);
+
+namespace Rollbar;
 
 use Throwable;
 use Psr\Log\AbstractLogger;
--- a/src/RollbarLogger.php
+++ b/src/RollbarLogger.php
@@ -86,7 +86,9 @@ class RollbarLogger extends AbstractLogger
     }
 
     /**
-     * @var Level|string $level
+     * @param Level|string $level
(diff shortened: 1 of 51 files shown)


src/RollbarLogger.php
```diff
@@ -163,12 +177,12 @@ public function shouldIgnoreError($errno)
...
     }
 
-    public function getQueueSize()
+    public function getQueueSize(): int
     {
         return count($this->queue);
     }
 
-    protected function send(\Rollbar\Payload\EncodedPayload $payload, $accessToken)
+    protected function send(EncodedPayload $payload, $accessToken): Response
```
waltjones, line 185: If I've read correctly, `string $accessToken` isn't allowed until PHP 7+, and that's why it's not used here and in the other function signatures that have a string argument?

waltjones (approved):
:+1: Looks good. I had one curiosity question about string types.

src/RollbarLogger.php
bishopb: Good catch, @waltjones! I had those changes stashed and forgot to pop them before I made this PR. I've addressed them in an addendum PR #524.

PR #522, opened 2021-05-25
fix: missing URL encoding of UUID in occurrences

bishopb:
## Description of the change

A URI should always be encoded, even when the expectation is that only conforming characters will be sent. In particular, the input for UUID comes from `src/DataBuilder.php` for both Agent and Fluent sender, and that is a UUID v4 value -- all characters of which conform to [RFC 3986][1]. A future code change might change that to a non-conforming string, or a consuming library might call the method directly with a non-conforming value. Be safe.

[1]:https://datatracker.ietf.org/doc/html/rfc3986

## Type of change

## Related issues

None

### Development

- [X] Lint rules pass locally

### Code review 

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -37,7 +37,7 @@ class Response
         if (!$this->wasSuccessful()) {
             return null;
         }
-        return "https://rollbar.com/occurrence/uuid/?uuid=" . $this->uuid;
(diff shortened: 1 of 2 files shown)


waltjones (approved):
:+1:

PR #526, opened 2021-05-28
refac: remove unreachable version compare branches

bishopb:
## Description of the change

No need to support code paths for versions less than PHP 8, since this code will not install or parse on versions lower than 8.

## Type of change

## Related issues

CH-79868

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/tests/DataBuilderTest.php
+++ b/tests/DataBuilderTest.php
@@ -804,9 +804,6 @@ class DataBuilderTest extends BaseRollbarTest
         $requestBody = $output->getRequest()->getBody();
         
         $this->assertEquals($streamInput, $requestBody);
-        if (version_compare(PHP_VERSION, '5.6.0') < 0) {
-            $this->assertEquals($streamInput, $_SERVER['php://input']);
-        }
         
         stream_wrapper_restore("php");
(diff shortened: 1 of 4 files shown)


tests/DataBuilderTest.php
```diff
@@ -892,7 +889,11 @@ public function testFramesOrder()
             $frames[count($frames)-1]->getFilename()
         );
         // 889 is the line number where the comment "// A" is found
-        $this->assertEquals(889, $frames[count($frames)-1]->getLineno());
+        $this->assertEquals(
+            886,
+            $frames[count($frames)-1]->getLineno(),
+            "Possible false negative: did this file change? Check the line number for line with '// A' comment"
+        );
```
bishopb, line 896: This test uses the test file itself as the input source for frames. Consequently, when the test file changes, the test may fail. The change here adds a comment, which will be output to the test runner, that the test failure may be false and should be cross-checked against the line number. Before, this same information was in the file itself as only a comment: now we also bubble this up to the test failure itself for greater visibility.

akornich (approved):
lgtm

PR #529, opened 2021-06-02
Remove throwable anachronism

bishopb:
## Description of the change

Since 3.0 now only supports PHP 8+, we can rely on `Throwable` as a base class for `Exception`. Anywhere we have BC logic catching one or the other should catch only `Throwable`.

## Type of change
- [X] New feature (non-breaking change that adds functionality)

## Related issues

Part of the CH-79868 series.

### Development

- [X] Lint rules pass locally

### Code review 

### DIFF
--- a/src/DataBuilderInterface.php
+++ b/src/DataBuilderInterface.php
@@ -2,7 +2,10 @@
 
 namespace Rollbar;
 
+use Rollbar\Payload\Data;
+use Throwable;
+
 interface DataBuilderInterface
 {
-    public function makeData($level, $toLog, $context);
+    public function makeData(string $level, Throwable|string $toLog, array $context): Data;
 }
--- a/src/Config.php
+++ b/src/Config.php
@@ -833,7 +833,7 @@ class Config
                     $this->verboseLogger()->info('Occurrence ignored due to custom check_ignore logic');
                     return true;
                 }
-            } catch (\Exception $exception) {
+            } catch (Throwable $exception) {
                 $this->verboseLogger()->error(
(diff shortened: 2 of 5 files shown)


src/DataBuilderInterface.php
```diff
@@ -2,7 +2,10 @@
 
 namespace Rollbar;
 
+use Rollbar\Payload\Data;
+use Throwable;
+
 interface DataBuilderInterface
 {
-    public function makeData($level, $toLog, $context);
+    public function makeData(string $level, Throwable|string $toLog, array $context): Data;
```
akornich, line 10: assuming "level" implies a Rollbar log level, can it be modeled as an enum type (if PHP supports such types)?

akornich (approved):
LGTM. It would be nice to eventually model log levels as an enum (instead of string) if possible. But that would change public API and would qualify for a new major release.

src/DataBuilderInterface.php
bishopb: @akornich - Interestingly, that model already exists: [`Rollbar\Payload\Level`][1]. API on the perimeter explicitly accept [`Level|string`][2], while internal "friend" API do so implicitly. [I commented][3] some while ago that this implicit type allowance isn't ideal, and that internally we should stick with `Level`.

[1]:https://github.com/rollbar/rollbar-php/blob/master/src/Payload/Level.php
[2]:https://github.com/rollbar/rollbar-php/blob/master/src/RollbarLogger.php#L92
[3]:https://github.com/rollbar/rollbar-php/blob/master/src/RollbarLogger.php#L99-L103

PR #536, opened 2021-08-02
Allow psr/log 2

Jean85:
## Description of the change
This PR allows the installation of `psr/log` 2.0 and 3.0. It's possible to support all of those versions at once since the interfaces are used but not implemented anywhere.
Backporting is unnecessary since both 2.0 and 3.0 require PHP 8.

This PR is also a blocker for Seldaek/monolog#1564

More details are available in the new meta document: https://www.php-fig.org/psr/psr-3/meta/

## Type of change

### Development
- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
   "require": {
     "php": ">=8.0.0 <9.0",
     "ext-curl": "*",
-    "psr/log": "^1 || ^2 || ^3",
+    "psr/log": "^1 || ^2",
     "monolog/monolog": "^2"
   },
 
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
   "require": {
     "php": ">=8.0.0 <9.0",
     "ext-curl": "*",
-    "psr/log": "^1",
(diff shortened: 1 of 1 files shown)


nicolas-grekas (dismissed):
(a quick merge would be appreciate as that creates some DX issues with new Symfony projects)

bxsx:
Hey @Jean85 and @nicolas-grekas.

It looks like our CI tests have failed. Does it work on your local machine?

Jean85:
I made a mistake. The interface is used through the AbstractLogger, so you can't support 3.0 without adding the void return type to all the methods.
This would technically be a breaking change on your users, since they would have the same breakage you had in CI if they extended your classes without the void return types.

bxsx:
Thanks for the clarification, @Jean85

In the current situation, I can merge the current PR (for 2.0) and put 3.0 request to our internal backlog so we can update the code in the next major release.

Jean85:
Great! Thank you!

bxsx (approved):

PR #539, opened 2021-10-21
Fix isset check

trsteel88:
```php
if (!isset($this->reportSuppressed)) {
    $this->reportSuppressed = isset($config['report_suppressed']) && $config['report_suppressed'];
}
```

The above line is never fired as ```$this->reportSuppressed = isset($config['reportSuppressed']) && $config['reportSuppressed'];``` is going to make the value set to true/false.

The code should actually check the array to see if the value isset.


bxsx (approved):
Hey @trsteel88 ! Good catch and thanks for the fix! :)

PR #534, opened 2021-07-05
fix: Unqualified Throwable refers to wrong namespace

edudobay:
## Description of the change

I found that `Rollbar::logUncaught`, added in 3.0, has an invalid signature — the unqualified `Throwable` refers to the nonexistent `Rollbar\Throwable`. This PR fixes that.

I also found an apparent mistake in that same method, where there was an unreachable statement after `return`. I moved the `return` after that statement. Please let me know if that should not be in this same PR. (In a separate commit)

## Type of change
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related issues

None — I'm reporting the bug and providing the fix in one go.

### Development

- [x] Lint rules pass locally
- [x] The code changed/added as part of this pull request has been covered with tests
- [x] All tests related to the changed code pass in development

### Code review 

- [ ] This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [ ] Changes have been reviewed by at least one other engineer
- [ ] Issue from task tracker has a link to this pull request

PR #547, opened 2021-11-14
fix: missed cleanup of synthetic member variable

bishopb:
## Description of the change

Release 3.0.0 added a `logUncaught` method to bypass the need to extend the PSR-defined `log` method. That implementation tagged the incoming object as "uncaught" and passed it through to the log method, which then checked for that tagging. However, `logUncaught` signature was incorrect and the pass through was not properly cleaning up the tagging and there weren't sufficient test coverage explicitly on the mechanism.

This elaboration on PR #534 adds that coverage.

## Type of change

## Related issues

[PR #534](https://github.com/rollbar/rollbar-php/pull/534)

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/tests/RollbarTest.php
+++ b/tests/RollbarTest.php
@@ -251,4 +251,23 @@ class RollbarTest extends BaseRollbarTest
         Rollbar::configure(array('enabled' => false));
         $this->assertTrue(Rollbar::disabled());
     }
+
+    public function testLogUncaughtUnsetLogger()
+    {
+        $sut = new Rollbar;
+        $result = $sut->logUncaught('level', new \Exception);
+        $expected = new Response(0, "Rollbar Not Initialized");
+        $this->assertEquals($expected, $result);
+    }
+
+    public function testLogUncaught()
+    {
+        $sut = new Rollbar;
+        Rollbar::init(self::$simpleConfig);
+        $logger = Rollbar::logger();
+        $toLog = new \Exception;
+        $result = Rollbar::logUncaught(Level::ERROR, $toLog);
+        $this->assertEquals(200, $result->getStatus());
+        $this->assertObjectNotHasAttribute('uncaught', $toLog);
+    }
 }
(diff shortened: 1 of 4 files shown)


tests/RollbarTest.php
```diff
@@ -251,4 +251,23 @@ public function testEnable()
...
+    public function testLogUncaught()
+    {
+        $sut = new Rollbar;
+        Rollbar::init(self::$simpleConfig);
+        $logger = Rollbar::logger();
+        $toLog = new \Exception;
+        $result = Rollbar::logUncaught(Level::ERROR, $toLog);
+        $this->assertEquals(200, $result->getStatus());
+        $this->assertObjectNotHasAttribute('uncaught', $toLog);
+    }
```
bxsx, line 272: `uncaught` is unset before and after invoking `logUncaught()`

The following assertion
```php
$this->assertObjectNotHasAttribute('uncaught', $toLog);
```
verifies the state after invocation but it should also verify the assumption of the unset variable before calling `logUncaught()` and verify that the state has changed inside the function.

bxsx (approved):
The fix looks fine but I have some concerns about provided tests that can be resolved before merging. Not a blocker though.

tests/RollbarTest.php
bishopb: I considered adding that test, @bxsx, but opted not to because `logUncaught` does _not_ have a pre-condition on the presence or value of `$toLog::uncaught`. However, `logUncaught` _does_ have a _post-condition_ that `$toLog::uncaught` is unset.

That said, we should refactor all of this in the future: there is a [proposal to deprecate dynamic properties on objects in 8.2][1] (on which I voted "yes") that will require a different approach should it pass.

[1]:https://wiki.php.net/rfc/deprecate_dynamic_properties

bxsx: Fair point! Thanks for the elaboration.

Issue #548, opened 2021-11-23
Uncaught TypeError: Rollbar\Payload\Frame::__construct(): Argument #1 ($filename) must be of type string


thomasme:
When logging a message (any level) with explicit Rollbar::log call, this error happens.

On PHP 8.0.12 with both 3.0.0 and master@ad22e3589116ff5958c3b6ff35c65d959266414d and Rollbar defaults.

Does not seem to happen when issues are caught through default error handlers.

Does not happen when `capture_error_stacktraces` is set to `false` on init.

```
<b>Fatal error</b>:  Uncaught TypeError: Rollbar\Payload\Frame::__construct(): Argument #1 ($filename) must be of type string, null given, called in /srv/vendor/rollbar/rollbar/src/DataBuilder.php on line 479 and defined in /srv/vendor/rollbar/rollbar/src/Payload/Frame.php:18
Stack trace:
#0 /srv/vendor/rollbar/rollbar/src/DataBuilder.php(479): Rollbar\Payload\Frame-&gt;__construct(NULL)
#1 /srv/vendor/rollbar/rollbar/src/DataBuilder.php(454): Rollbar\DataBuilder-&gt;makeFrames(Object(TypeError), false)
#2 /srv/vendor/rollbar/rollbar/src/DataBuilder.php(425): Rollbar\DataBuilder-&gt;makeTrace(Object(TypeError), false)
#3 /srv/vendor/rollbar/rollbar/src/DataBuilder.php(406): Rollbar\DataBuilder-&gt;getExceptionTrace(Object(TypeError))
#4 /srv/vendor/rollbar/rollbar/src/DataBuilder.php(375): Rollbar\DataBuilder-&gt;getBody(Object(TypeError), Array)
#5 /srv/vendor/rollbar/rollbar/src/Config.php(732): Rollbar\DataBuilder-&gt;makeData('...', Object(TypeError), Array)
#6 /srv/vendor/rollbar/rollbar/src/RollbarLogger.php(215): Rollbar\Config-&gt;getRollbarData('...', Object(TypeError), Array)
#7 /srv/vendor/rollbar/rollbar/src/RollbarLogger.php(120): Rollbar\RollbarLogger-&gt;getPayload('...', '...', Object(TypeError), Array)
#8 /srv/vendor/rollbar/rollbar/src/Handlers/ExceptionHandler.php(28): Rollbar\RollbarLogger-&gt;log('...', Object(TypeError), Array)
[...]

PR #78, opened 2016-03-24
Add me alwaysontop617

alwaysontop617:
I want to have my name in the Thanks file.

PR #552, opened 2021-12-03
fix: psalm #78, possibly null argument

bishopb:
## Description of the change

Psalm reports numerous instances of possibly null arguments. One of which was reported as Issue #548, a legitimate bug. While most of the other instances are code hygiene, another was also a bug in IP address handling.

Note that this PR is based upon the branch of PR #551.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fixes Issue #548 

### Development

- [X] Lint rules pass locally

### Code review

### DIFF
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -1218,9 +1218,9 @@ class DataBuilder implements DataBuilderInterface
         try {
             if (function_exists('shell_exec')) {
                 $stdRedirCmd = Utilities::isWindows() ? " > NUL" : " 2> /dev/null";
-                $output = rtrim(shell_exec('git rev-parse --abbrev-ref HEAD' . $stdRedirCmd));
+                $output = shell_exec('git rev-parse --abbrev-ref HEAD' . $stdRedirCmd);
                 if ($output) {
-                    return $output;
+                    return rtrim($output);
                 }
             }
             return null;
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -549,18 +549,25 @@ class DataBuilder implements DataBuilderInterface
 
     protected function getLevel($level, $toLog)
     {
-        if (is_null($level)) {
+        // resolve null level to default values, if we can
+        if ($level === null) {
             if ($toLog instanceof ErrorWrapper) {
-                $level = isset($this->errorLevels[$toLog->errorLevel]) ? $this->errorLevels[$toLog->errorLevel] : null;
+                $level = $this->errorLevels[$toLog->errorLevel] ?? null;
             } elseif ($toLog instanceof \Exception) {
                 $level = $this->exceptionLevel;
             } else {
                 $level = $this->messageLevel;
             }
         }
-        $level = strtolower($level);
-        $level = isset($this->psrLevels[$level]) ? $this->psrLevels[$level] : null;
-        return $this->levelFactory->fromName($level);
+        if ($level !== null) {
+            $level = strtolower($level);
+            $level = $this->psrLevels[$level] ?? null;
+            if ($level !== null) {
+                // this is a well-known PSR level: "error", "notice", "info", etc.
+                return $this->levelFactory->fromName($level);
+            }
+        }
+        return null;
     }
 
     protected function getTimestamp()
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -468,13 +468,15 @@ class DataBuilder implements DataBuilderInterface
         $frames = array();
         
         foreach ($this->getTrace($exception) as $frameInfo) {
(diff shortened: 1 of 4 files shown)


src/DataBuilder.php
```diff
@@ -818,27 +830,35 @@ protected function getRequestBody()
         return $this->requestBody;
     }
 
-    /*
+    /**
+     * Get the user's IP, by inspecting the http header X-Real-IP, or if not
+     * that first address from the http header X-Forwarded-For, and if not that
+     * then the remote IP connecting to the web server, if available.
+     *
```
bxsx, line 837: Other SDKs (e.g. Pyrollbar, Gem) handle it in the opposite way with X-Forwarded-For checking first.

bxsx (approved):
LGTM
Leaving one note regarding SDKs consistency.

src/DataBuilder.php
bishopb: @bxsx Would you like this to switched? If so, I recommend a separate task and prioritization.

src/DataBuilder.php
```diff
@@ -1193,19 +1213,15 @@ public function detectGitBranch($allowExec = true)
...
+            $stdRedirCmd = Utilities::isWindows() ? ' > NUL' : ' 2> /dev/null';
+            $output = shell_exec('git rev-parse --abbrev-ref HEAD' . $stdRedirCmd);
+            if (is_string($output)) {
+                return rtrim($output);
             }
-            return null;
-        } catch (\Exception $e) {
-            return null;
         }
+        return null;
```
bxsx, line 1225: I think this can be simplified after refactoring a null logic:

```php
    private static function getGitBranch(): ?string
    {
        if (!function_exists('shell_exec')) return null;

        $stdRedirCmd = Utilities::isWindows() ? ' > NUL' : ' 2> /dev/null';
        $output = shell_exec('git rev-parse --abbrev-ref HEAD' . $stdRedirCmd);
        if (!$output) return null;
        return rtrim($output);
    }
```

Regarding removing try/catch block - we don't need to clean this up in case of occurring `E_WARNING` for `shell_exec`, correct?

bishopb: I didn't want to stray too far from the style of the original code. The changes here are meant to be obviously faithful to the original so as to not obscure the intent of my change (namely fixing psalm issues, not style). That said, the suggestion doesn't pass the linting rules:
```
FILE: /opt/rollbar/rollbar-php/src/DataBuilder.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 1218 | ERROR | [x] Inline control structures are not allowed
 1222 | ERROR | [x] Inline control structures are not allowed
```

> Regarding removing try/catch block - we don't need to clean this up in case of occurring E_WARNING for shell_exec, correct?

Not in this changeset, no. There are plenty of places where the code ignores the raised errors, polutting the users logs. Those need a pass.

bxsx: It's not about the style of the code. It's all about the wrong design of the logic. If the original code is of bad quality, I don't think we should reuse the wrong patterns.

Your implementation does not provide a clean solution for possibly null arguments. The code doesn't self-document, provide rationales for the next steps, etc. It is difficult to read/analyze and may reduce maintainability in the future. The code I suggest follows the design pattern of composing method.

> That said, the suggestion doesn't pass the linting rules:

I don't know the lining rules but the output clearly shows how to fix it:

```php
    private static function getGitBranch(): ?string
    {
        if (!function_exists('shell_exec')) {
            return null;
        }

        $stdRedirCmd = Utilities::isWindows() ? ' > NUL' : ' 2> /dev/null';
        $output = shell_exec('git rev-parse --abbrev-ref HEAD' . $stdRedirCmd);
        if (!$output) {
            return null;
        }
        return rtrim($output);
    }
```

src/DataBuilder.php
```diff
@@ -818,27 +830,35 @@ protected function getRequestBody()
...
             $parts = explode(',', $forwardFor);
             $ipAddress = $parts[0];
         }
-        $realIp = isset($_SERVER['HTTP_X_REAL_IP']) ? $_SERVER['HTTP_X_REAL_IP'] : null;
+        $realIp = $_SERVER['HTTP_X_REAL_IP'] ?? null;
         if ($realIp) {
             $ipAddress = $realIp;
         }
+
+        if (null === $ipAddress) {
```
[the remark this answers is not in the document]
bxsx: Has it been accidentally marked as resolved? I "unresolve" it for now.

jmarronm:
Let´s merge this PR and create new tickets to track separately SDK consistency and additional comments
Thanks!

jmarronm (approved):
Approved to merged pending of capturing additional comments in dev tickets

PR #554, opened 2021-12-27
Possibility to define custom scrubber

### LINKED ISSUE #477: Scrubbing HTTP_COOKIE

zinkovskiy:
## Description of the change

In current implementation scrubbing fields is not flexible.
Sometimes we need to scrub particular cookie value without erasing all cookies, like suggested here: #477
So, why can't we add an option to define custom scrubber, especially when all the code has been written before? :slightly_smiling_face: 

## Type of change

## Related issues

- Fix #477

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -58,6 +58,7 @@ class Config
         'capture_username',
         'capture_email',
         'root',
+        'scrubber',
         'scrub_fields',
         'scrub_safelist',
         'timeout',


danielmorell:
Hi @zinkovskiy, sorry for the long wait on getting some feedback on this!

Rollbar already supports the ability to create a custom scrubber. The `scrubber` configuration name is already used to define a custom scrubber class. The class must implement the `Rollbar\ScrubberInterface` interface. I would recommend extending the `Rollbar\Scrubber` class as the easiest way to get up and going quickly.

Adding `'scrubber'` to the list of options does not enable it, nor change the behavior of the SDK. Because of that I am going to close this PR.

zinkovskiy:
> Hi @zinkovskiy, sorry for the long wait on getting some feedback on this!
> 
> Rollbar already supports the ability to create a custom scrubber. The `scrubber` configuration name is already used to define a custom scrubber class. The class must implement the `Rollbar\ScrubberInterface` interface. I would recommend extending the `Rollbar\Scrubber` class as the easiest way to get up and going quickly.
> 
> Adding `'scrubber'` to the list of options does not enable it, nor change the behavior of the SDK. Because of that I am going to close this PR.

Hi @danielmorell 
I insist that `scrubber` should be in the list of options
[Symfony bundle](https://packagist.org/packages/rollbar/rollbar-php-symfony-bundle) is used to integrate rollbar with symfony apps.
The bundle defines convenient way to configure rollbar.
The bundle obtain available configuration options using [\Rollbar\Config::listOptions](https://github.com/rollbar/rollbar-php/blob/master/src/Config.php#L258) method, that just returns static array $options field.

`scrubber` is missing in this array, due to this the bundle does not allow to set `scrubber` option when configure rollbar
Error `Unrecognized option "scrubber" under "rollbar". Available options are <...>` will be shown when `cache:clear` symfony command is called

So, could you please get back to this issue and take a look on it one more time?

P. s.
I can create repository with example of symfony application with rollbar bundle if you need example of the bug.

danielmorell (approved):
Hi @zinkovskiy, I appreciate your persistence on this, and apologize for overlooking the use case you have identified. Thank you for not just giving up! Based on what you have said, I now think this is an important change and happily approve it.

zinkovskiy:
> Hi @zinkovskiy, I appreciate your persistence on this, and apologize for overlooking the use case you have identified. Thank you for not just giving up! Based on what you have said, I now think this is an important change and happily approve it.

Hi @danielmorell, thank you :pray: Do you have estimation when these changes will be released?

Issue #559, opened 2022-03-11
Rollbar\\Payload\\Body::__construct(): Argument #1 ($value) must be of type Rollbar\\Payload\\ContentInterface, Rollbar\\Payload\\TraceChain given


stephpy:
On Rollbar-php 3.1.0:

https://github.com/rollbar/rollbar-php/blob/master/src/DataBuilder.php#L409

Fail when getExceptionTrace returns a TraceChain since TraceChain has not the ContentInterface.
PR without BC break is coming.

danielmorell:
Thank you for bringing this to our attention!

PR #560, opened 2022-03-11
Tracechain must implements ContentInterface to be part of Body.

stephpy:
## Description of the change

In current release (3.1.0), When TraceChain is given to Body, an exception is given because TraceChain does not implements ContentInterface.

## Type of change

## Related issues

- Fix https://github.com/rollbar/rollbar-php/issues/559

### DIFF
--- a/src/Payload/TraceChain.php
+++ b/src/Payload/TraceChain.php
@@ -4,7 +4,7 @@ namespace Rollbar\Payload;
 
 use Rollbar\SerializerInterface;
 
-class TraceChain implements SerializerInterface
+class TraceChain implements ContentInterface
 {
     public function __construct(private array $traces)
     {


danielmorell (changes requested):
Thank you for your submitting this update! Could we make one small change? Let's also include the `SerializerInterface`. Otherwise we will be running to some deprecation warnings.

```
class TraceChain implements SerializerInterface, ContentInterface
```

Could you add a new commit that includes both interfaces? Thank You!

danielmorell (approved):
Never mind my previous comment. I'm an idiot. I forgot `ContentInterface` extends `SerializerInterface`. 

Your update is perfect the way it is!

stephpy:
You're welcome :) 🍻

PR #565, opened 2022-04-12
Fix error suppressor context detection for PHP 8 

tanakahisateru:
## Description of the change

I upgraded PHP and Rollber client then too many errors are reported. I'm using `@` operator.
PHP 8 changed `error_reporting()` return value when it used with `@`.

https://www.php.net/manual/en/language.operators.errorcontrol.php

```php
<?php
function f() {
    echo error_reporting(), PHP_EOL;
}

error_reporting (E_ALL);
f();    // -1
@f();   // PHP7:0 / PHP8:4437
```

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Maintenance
- [ ] New release

## Related issues

> Shortcut stories and GitHub issues (delete irrelevant)

- Fix [SC-]
- Fix #1

### Development

- [x] Lint rules pass locally
- [x] The code changed/added as part of this pull request has been covered with tests
- [x] All tests related to the changed code pass in development

### Code review

- [ ] This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ] "Ready for review" label attached to the PR and reviewers assigned
- [ ] Issue from task tracker has a link to this pull request
- [ ] Changes have been reviewed by at least one other engineer

PR #566, opened 2022-04-12
Safer parse_str() usage

tanakahisateru:
## Description of the change

`parse_str()` reports warning if parsed items exceeded over than`max_input_vars`. Rollbar Logger logs this internal error instead of user errors.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

None

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/DataBuilder.php
+++ b/src/DataBuilder.php
@@ -178,7 +178,7 @@ class DataBuilder implements DataBuilderInterface
         $fromConfig = isset($config['local_vars_dump']) ? $config['local_vars_dump'] : null;
         $this->localVarsDump = self::$defaults->localVarsDump($fromConfig);
         if ($this->localVarsDump && !empty(ini_get('zend.exception_ignore_args'))) {
-            ini_set('zend.exception_ignore_args', 'Off');
+            ini_set('zend.exception_ignore_args', '0');
             assert(empty(ini_get('zend.exception_ignore_args')));
         }
     }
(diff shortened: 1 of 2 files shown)


danielmorell:
I just read #565, which adds a little more context about your thought process. Can you think of a scenario where we would have an error generated from `parse_str()` inside the Rollbar SDK that would not be generated elsewhere as well?

tanakahisateru:
Thanks @danielmorell . I actually experienced that no original errors reported while 100+ of `parse_str()` warnings.

```php
function badTypeFunction(string $str): string
{
    return preg_replace(..., $str, ...); // returns null when internal error occurred.
}
```

We got a massive data violation to this function like below:

```
&#3604;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;&#3655;... (more than 1000 strange pattern)
```

tanakahisateru:
I think that PHP native functions don't have proper error handling. For example, `fopen()` and `fclose()` should be used with `@` and return value for unexpected IO error. Though ignoring items after 1001 is bad idea, `parse_str()` never do so correctly even if `@` removed. This is not general better practice but special case for this PHP behavior. (This is just my opinion. It's OK that someone else don't think same.)

tanakahisateru:
My another idea is like below:

```php
final class Utilities
{
    public static function saferParseStr($string)
    {
        try {
            parse_str($string, $result);
        } catch (\Throwable) {
            @parse_str($string, $result);
            $result = array_merge($result, ["...more than max_input_vars items"]);
        }
        return $result;
    }
}
```

danielmorell (approved):
Long term it would be nice to avoid running every string through `parse_str()`, but that would take some rethinking of how we handle inputs. 

I think this is a good solution for our current problem. Thank you @tanakahisateru!

PR #568, opened 2022-05-03
tests: fix running tests on local machine (mac m1)

Chris53897:
## Description of the change

The Value "Off" does not solve to true for empty() on my local machine to run the tests.
I see that it works on github actions.
This fix solves it for my case.

## Type of change


danielmorell (approved):
This looks like a good addition. The values we want to see would be `false`, `"0"`, or `"Off"`. All of which are valid (and possible) responses from `ini_get()`.

danielmorell (approved):

Issue #461, opened 2019-06-03
Respect exception context in RollbarLogger


rieschl:
When using the PSR-3 `LoggerInterface` implementation `\Rollbar\RollbarLogger` directly in a class to log certain exceptions, Rollbar doesn't respect the `exception` array key in the `context` parameter, the exception has to be passed as `message` to see the pretty stack trace in the Rollbar web interface.

Although technically this is ok, because PSR-3 states that `Every method accepts a string as the message, or an object with a __toString() method.` and  the `Exception` (and `Throwable`) class has a `__toString()` method, it would be nice to also be able to pass the exception via the `$context['exception']` key and use a string as `$message`. PSR-3 [defines that behavior explicitly](https://www.php-fig.org/psr/psr-3/#13-context).

So the following code snippets should log the same thing in Rollbar:

```
/**
 * @var \Psr\Log\LoggerInterface $logger
 * @var \Throwable $e
 */
$logger->error($e); // current
$logger->error($e->getMessage(), ['exception' => $e]); // suggested
```

Thanks :)

danielmorell:
This has been resolved and will be releases in the new `v4.0.0` release.

PR #577, opened 2022-10-24
Added support for psr/log v3

### LINKED ISSUE #570: Plans for psr/log v.3?
Enhancement:

Hi,
Are there any plans to migrate from psr/log from v2 to v3 ?
I did a quick look into the code, and i think it is not so easy to change it.

[actual](https://github.com/rollbar/rollbar-php/blob/master/src/RollbarLogger.php#L92):
`#public function log($level, $toLog, array $context = array())`

[psr/log v.3](https://github.com/php-fig/log/blob/master/src/LoggerInterface.php#L124)
`public function log($level, string|\Stringable $message, array $context = []): void`

Problem 1 the return instead of void.
Problem 2 $toLog has a different type.

Maybe a solution could be to remove psr/log in the long term?
So usere are not tied to v1|v2

Greetz

danielmorell:
## Description of the change

This PR adds support for `psr/log` version 3. This has been a long-standing issue but could not be fixed without causing breaking changes. See #536 for a discussion.

I renamed the `log()` method in both `Rollbar` and `RollbarLogger` classes to `report()`. New `log()` methods were then created that were compliant with `Psr\Log\LoggerInterface`. Since the `report()` methods are not defined in the `psr/log` interface we can return the `Response` object. This is very important for testing purposes.

Almost all unit tests that used `Rollbar::log()` or `RollbarLogger::log()` were updated to use `Rollbar::report()` or `RollbarLogger::report()`. There are a few exceptions where the `log()` method was still tested. To validate that `log()` was working correctly a test helper class `ArrayLogger` was created that collects all verbose logs into an array in memory. Those logs can then be checked as a side effect of calling `log()` since it no longer returns anything.

It is worth noting that there is still one outstanding behavior defined by `psr/log` that has not been implemented. #461 explains this. I will resolve that issue in a separate PR.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix #570

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -146,8 +146,11 @@ class Rollbar
             return self::getNotInitializedResponse();
         }
         $toLog->isUncaught = true;
-        $result = self::$logger->report($level, $toLog, (array)$extra);
-        unset($toLog->isUncaught);
+        try {
+            $result = self::$logger->report($level, $toLog, (array)$extra);
+        } finally {
+            unset($toLog->isUncaught);
+        }
         return $result;
     }
     
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -2,10 +2,12 @@
 
 namespace Rollbar;
 
+use Psr\Log\InvalidArgumentException;
 use Rollbar\Payload\Level;
 use Rollbar\Handlers\FatalHandler;
 use Rollbar\Handlers\ErrorHandler;
 use Rollbar\Handlers\ExceptionHandler;
+use Stringable;
 use Throwable;
 
 class Rollbar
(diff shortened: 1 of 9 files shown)


src/Rollbar.php
```diff
@@ -90,12 +92,49 @@ public static function scope($config)
...
+     *                                   the {@see Level} constants.
+     * @param string|Stringable $message The log message.
+     * @param array             $context Arbitrary data.
+     *
+     * @return void
+     *
+     * @throws InvalidArgumentException If $level is not a valid level.
+     * @throws Throwable Rethrown $message if it is {@see Throwable} and {@see Config::raiseOnError} is true.
+     */
+    public static function log($level, string|Stringable $message, array $context = array()): void
```
anler, line 109: I guess the reason we don't type the `$level` param here is because of the `psr/log` interface we want to abide, right?

src/Rollbar.php
```diff
@@ -107,49 +146,49 @@ public static function logUncaught($level, Throwable $toLog, $extra = array())
             return self::getNotInitializedResponse();
         }
         $toLog->isUncaught = true;
-        $result = self::$logger->log($level, $toLog, (array)$extra);
+        $result = self::$logger->report($level, $toLog, (array)$extra);
         unset($toLog->isUncaught);
```
anler, line 150: What do you think about wrapping this block in a try-finally to make sure we always unset `isUncaught`?

src/Rollbar.php
```diff
@@ -247,7 +286,7 @@ public static function destroy()
     public static function report_exception($exc, $extra_data = null, $payload_data = null)
```
anler, line 286: Does it make sense to annotate the return type here as we're doing with the rest of methods?

tests/TestHelpers/ArrayLogger.php
```diff
@@ -0,0 +1,58 @@
...
+use Psr\Log\AbstractLogger;
+use Stringable;
+
+
+/**
+ * Simple logger that saves logs to an array.
+ *
+ * @since 4.0.0
+ */
+class ArrayLogger extends AbstractLogger
```
anler, line 14: Nice! ✨

anler (approved):
Only minor comments, good work!

src/Rollbar.php
danielmorell: That is correct. `psr/log` does not have a type annotation for `$level` so adding one would break the interface implementation.

src/Rollbar.php
danielmorell: That is a really solid idea. Otherwise, it would be possible for our dynamic property to persisting if the `Throwable` was thrown in `report()`.

I have committed that change.

src/Rollbar.php
danielmorell: I am planning on adding type annotations and DocBlocks to most everything. I will submit that in a separate PR.

PR #578, opened 2022-10-31
Removed deprecated log levels and fixed inconsistent use of LevelFactory

danielmorell:
### 1. Removed deprecated code from `Level`
The first major change removes some deprecated constants and the `__callStatic()` from the `Rollbar\Payload\Level` class. These have been deprecated since `v1.2.0`, and it seems like now is a good time to remove them. This means that, although deprecated, a `Level` instance could be constructed by calling `Level::ERROR()`. This was very concise but lacked any type annotations or consistency. The `LevelTest.php` file shows this.

### 2. Refactored `LevelFactory` with static methods
The second major change is changing the public `LevelFactory` methods to be static and replacing the `LevelFactory::init()` method with the `LevelFactory::getLevels()` method. This change eliminates the need to instantiate `LevelFactory` before calling any methods. Creating a valid new `Level` instance is more verbose than I like. Making the methods static reduces the total code needed to validate and create a new `Level`. See the examples below.

```php
// v1.2.0
$levelObject = Level::ERROR();
```

```php
// v1.2.1+
$levelFactory = new LevelFactory();
$levelObject = $levelFactory->fromName(Level::ERROR);
```

```php
// v4.0.0
$levelObject = LevelFactory::fromName(Level::ERROR);
```

Perhaps in a PHP 8.1 and above world we can use a backed `enum` and a few methods with `match` expressions to eliminate the need for the `LevelFactory` class entirely.

### 3. Replaced instances of `LevelFactory` with static method calls

The third significant change is removing the `$levelFactory` property from the `Rollbar\Config`, `Rollbar\DataBuilder`, and `Rollbar\RollbarLogger` classes. Since the `LevelFactory` methods can now be called statically, we no longer need instances of `LevelFactory` easily accessible on these classes.

This has the possibility to introduce a change in behavior. It was possible for someone to provide their own level factory class when instantiating `DataBuilder`. However, I don't think this would be likely or work very well since `RollbarLogger` attempts to validate the level using our `LevelFactory` class before it gets to `DataBuilder`.

With only partial support for a custom level factory theoretically available this feels like something that should be cleaned up. We should use one level factory across the codebase. Either provided by the user or our own. Support for a custom level factory is undocumented, partial, and probably just a bad idea anyway. Because of that I have completely removed it.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

None.

### Development

- [x] Lint rules pass locally

### Code review


cyrusradfar (approved):
gave feedback directly, thank you for cleaning this up and aligning us more w the standard

PR #580, opened 2022-11-17
Renamed IStrategy to StrategyInterface updated Truncation

danielmorell:
The previous `IStrategy` interface suffers from a few issues.

1. It does not fit our current naming convention. We use `<name>Interface` for all other interfaces while it uses `I<name>`.
2. The type definitions in the method PHPDoc were incorrect. For example the `@return` tag for `IStrategy::applies()` specifies the return type as `array`. However, all the classes that implement the interface a `bool` is returned. Additionally, a `bool` is clearly expected when the `applies()` method is called... https://github.com/rollbar/rollbar-php/blob/169cee9ef78139558d6a3e5f680afb5853f4228e/src/Truncation/Truncation.php#L49
3. The interface is missing any documentation explaining the expected behavior and use of the methods it describes.

## Description of the change

These core issues perpetuated themselves through the implementations of the `IStrategy` interface. To resolve all of this, I have done the following...

1. Renamed `IStrategy` to `StrategyInterface`.
2. Added type annotations to the interface method signatures.
3. Updated the PHPDoc for the interface methods so that the expected behavior is documented and the types match.
4. Updated the type annotations on all implementations of the `StrategyInterface` starting with the `AbstractStrategy`.

The updates to the interface would cause a BC since some user's custom truncation classes may not be complient with the interface anymore.

I also took the opportunity to clean up and document what each truncation strategy does and how payload truncation itself works. While doing so I also cleaned up some minor code quality issues. (This PR is mostly adding method PHPDocs). However, there are a few important code changes...

1. Custom truncation strategy validation was changed from requiring a custom strategy class extend the `AbstractStrategy` to now require it implement `StrategyInterface`. Any class that extends `AbstractStrategy` will still pass the validation.
2. `FramesStrategy::selectFrames()` is now a static method. This change was introduced because an instance of `FramesStrategy` was being instantiated by `MinBodyStrategy` just so `selectFrames()` could be called.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

None

### Development

- [x] Lint rules pass locally

### Code review


waltjones (approved):
:+1: Nice to see tech debt paid down.

PR #586, opened 2022-12-19
Added required public methods to the `DataBuilderInterface`

danielmorell:
## Description of the change

**The first major update** is adding public methods to `DataBuilderInterface` that are called during the normal course of operation. This is evidenced by the fact that the test fixture `FakeDataBuilder` included more methods than the interface required. If those methods were removed, the tests would start throwing errors because methods were called that did not exist.

This is important because it is possible for a user to do something like the following...

```php
Rollbar::init(['dataBuilder' => MyCustomDataBuilder::class]);
```

The `dataBuilder` config option is not readily apparent in the code and is not in our developer documentation. (I didn't realize it existed until a few months ago.)

If a user supplies their own data builder class, it must implement the `DataBuilderInterface`. However, if the user only implemented the methods that are required by the interface, we would start throwing exceptions when methods are called that don't exist. This honestly seems like a bit of an edge case, since rewriting the entire `DataBuilder` class seems unlikely. Most people would extend the `DataBuilder` class.

The reason this is being included is because fixing it is a breaking change. Otherwise, it seems like a minor issue.

**The second major change** is removing the tests around custom data that is not an array. Our entire logic around custom data assumes the 'custom' config key is an array. In March of this year, I added these tests to ensure primitives and different objects would work correctly. At the time I was not as aware of the rest of our codebase that deals with the custom data array. Interestingly these tests passed. However, trying to use something other than an array would cause issues elsewhere.

I also updated the logic in `DataBuilder::addCustom()` to remove the need to throw an exception.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

None

### Development

- [x] Lint rules pass locally

### Code review


waltjones (approved):
@danielmorell thank you for the thoughtful description and PR.

PR #589, opened 2022-12-25
chore: add ci tests for php 8.2, fix ci deprecations

Chris53897:
bump github action versions.

## Type of change

- [x] Chore CI

Fix deprecation of github actions

### DIFF
--- a/.github/workflows/ci-1.x.yml
+++ b/.github/workflows/ci-1.x.yml
@@ -29,9 +29,9 @@ jobs:
   # compatible version as well as the most-recent stable version. This will
   # fail-fast by default, so we include our edgiest versions (currently 7.4)
   # first as they're most likely to fail.
-  # @seealso https://freek.dev/1546
-  # @seealso https://www.dereuromark.de/2019/01/04/test-composer-dependencies-with-prefer-lowest/
-  # @seealso https://github.com/actions/starter-workflows/blob/main/ci/php.yml
+  # @see https://freek.dev/1546
+  # @see https://www.dereuromark.de/2019/01/04/test-composer-dependencies-with-prefer-lowest/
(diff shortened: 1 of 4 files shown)


Chris53897:
There is 1 error for the run in php 8.2 https://github.com/rollbar/rollbar-php/actions/runs/3775480588/jobs/6418187398
But this should be fixed in a seperate PR.

danielmorell (changes requested):
@Chris53897 thank you for the PR. I agree that we should address the PHP 8.2 compatibility issues in a separate PR. However, because those issues have not been addressed, we will have inconsistent test runs on PHP 8.0 and 8.1 since they may be stopped before completing if the PHP 8.2 tests fail first.

I can see two ways to resolve this:

1. Remove PHP 8.2 from this PR and add it as part of the 8.2 compatibility update. 
2. Add `fail-fast: false` to the `strategy` in the `ci.yml` file. This would ensure that all tests in the matrix are run even if one fails before another finishes.

Let me know what you think.

Chris53897:
@danielmorell Looks like this little change makes the tests pass. I just saw you opened an issue https://github.com/rollbar/rollbar-php/issues/590
I am not aware of side effects this little change can have. Please have a look.

danielmorell:
@Chris53897 I made that same change locally while investigating PHP 8.2 compatibility! I love it! And you are not wrong. That is all that is needed to make the tests pass. 

I need to look deeper into the `ExceptionHandler` tests. I expect we should have a deprecation error when trying to create the `isUncaught` property on an `Exception` instance. But that did not appear to show up in any of the tests.

danielmorell (approved):
I think we should fix the `ExceptionHandler` in a separate PR. So, this looks good! Thank you @Chris53897!

PR #593, opened 2023-01-10
Updated the `TransformerInterface`

danielmorell:
## Description of the change

In this PR I have added comments to the `TransformerInterface`.

I also updated the return type on the `TransformerInterface::transform()` method from a union of `null` and `Payload` to just be `Payload`. The reason for this is, a `null` return type has not been accounted for in the logic flow where the payload is used. It is expected to be a `Payload` instance not `null`. 

I cannot think of a reason to transform the payload to `null` other than to have the error ignored. However, check ignore is the very next step. So, that does not seem like plausible need either. 

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

None

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -795,7 +795,7 @@ class Config
         Level|string $level,
         mixed $toLog,
         array $context = array ()
-    ): ?Payload {
+    ): Payload {
         if (count($this->custom) > 0) {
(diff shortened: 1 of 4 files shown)


waltjones (approved):
Rationale to remove null makes sense. Thanks for the explanation.

PR #594, opened 2023-01-16
fix: correct version string

donoskaro:
## Description of the change

Currently, the Rollbar PHP library is identifying itself with an older version string causing a notice to appear in the Rollbar UI.

> rollbar-php version 3.1.4 is now available. This project uses version 3.1.2. [View Release Notes](https://github.com/rollbar/rollbar-php/releases)

This change updates the version number to the latest one plus one to account for SemVer.

## Type of change


danielmorell:
@donoskaro thank you for bringing this to our attention. Our next v3 release will be v3.2.0 as we are planning on deprecating some things to help people transition to the new v4.0.0 version.

Issue #405, opened 2018-08-30
Rollbar::init() shouldn't require access token when using agent as handler


joshrai:
In v0.18.2, it's legal and workable to call `Rollbar::init()` without passing an `access_token` entry in the `$config` param, if you do pass an entry of `'handler' => 'agent'`:

https://github.com/rollbar/rollbar-php/blob/fafe13f8beb669d6bba0137f3703fa808d50cbb7/src/rollbar.php#L190-L192

But in v1.6.2, and possibly other earlier versions, `Rollbar::init()` always requires an `access_token` entry, and furthermore requires it to be a string of length 32, because it always reaches this line:

https://github.com/rollbar/rollbar-php/blob/1deec89b870899d12794bb75377ca60359f18723/src/Config.php#L248

I think this should not be required if we're using the Rollbar agent as the handler, because this library shouldn't need to communicate with Rollbar's server in that case; it should just be writing events to a local log file that a separate agent process will pick up. I was advised by support that this might have been an oversight, and that using a dummy value should serve as a workaround. That seems a little messy and misleading; I think it would be better if the library detected when it does and doesn't need an access token.

I tried this out using HHVM 3.24.7, but it seems like the code path in question would be relevant for any PHP or HHVM version.

joshrai:
By the way, it looks like the workaround of using an arbitrary 32-char value for `access_token` doesn't work after all. Providing a dummy value does bypass the validation that the library does, but it seems that you need to pass a real access token to the client library in order for the agent to succeed. If I use a dummy value in the `Rollbar::init()` configuration, I see this error in the Rollbar agent log when I try to log something:

```
2018-10-08 23:34:57,327 WARNI Unexpected response from Rollbar API. Code: 401 Body: {
"err": 1,
"message": "invalid access token"
}
```

and log entries do not get through to the Rollbar service. If I use the correct access token in `Rollbar::init()`, then my log entries do get through to Rollbar, and I no longer see that error in the agent log.

My understanding is that you must [configure the agent itself](https://github.com/rollbar/rollbar-agent#configuration) to know the access token, which made me think that requiring it additionally in the client seemed redundant. Maybe there is a security rationale for requiring it in both places. But if so, it would be great for the `rollbar-php` docs to explain why. Thanks!

JohnKendhammer:
@joshrai You are very much right that this is not updated in the documentation and is something that I am updating.

joshrai:
[...]

PR #595, opened 2023-01-16
Fixed #469 Added `requireAccessToken()` method to senders

### LINKED ISSUE #469: Access token is required even if `handler` configuration parameter is set
- `rollbar/rollbar-php-symfony-bundle` v3.1.2
- `rollbar/rollbar` v1.7.4

We recently changed our Rollbar configuration so that it uses fluentd td-agent for logging errors. So we added :
```
handler: fluent
fluent_tag: identifier_for_component_logging_to_rollbar
```
to the configuration file and of course removed the `access_token` variable as it is no longer needed since this is now added to the configuration file of the td-agent daemon. However running `composer install` fails with:

```
In Utilities.php line 25:                                                                                                                                            
                                                                                                                                                                       
    $config['access_token'] must not be null                                                                                                                           
                                                                                                                                                                       
  In Utilities.php line 25:                                                                                                                                            
                                                                                                                                                                       
    $config['access_token'] must not be null                                                                                                                           
                                                                                                                                                                       
```

Adding a false access token resolves this issue, and of course Rollbar works as expected. 

Wouldn't it make more sense to check if fluent handler is used and if so, to not call the 

```
private function setAccessToken($config)
    {
        if (isset($_ENV['ROLLBAR_ACCESS_TOKEN']) && !isset($config['access_token'])) {
            $config['access_token'] = $_ENV['ROLLBAR_ACCESS_TOKEN'];
        }
        $this->utilities->validateString($config['access_token'], "config['access_token']", 32, false);
        $this->accessToken = $config['access_token'];
    }
```
method which in turn calls the validateString method that generates this error ?

danielmorell:
## Description of the change

We have three different built in payload senders that take a payload and send it to the Rollbar service. Some require that the payload include the access token. The `Rollbar\Senders\CurlSender` class is one of these. Our other two built in senders can use an external access token. The `Rollbar\Senders\FluentSender` is does not need the access token. And it was causing issues (see #469). 

This proposed change moves the requirement for knowing if the access token is required from the `Rollbar/Config` class to the configured sender. All senders must implement `Rollbar\Senders\SenderInterface`. So, I added a new `requireAccessToken()` method to the `SenderInterface`. The method returns a `boolean` indicating if the sender requires the access token. This method is called during the initialization process. If the access token is required and not valid an error is thrown.

The reason this is a breaking change is because it requires the `requireAccessToken()` method be added to all custom sender interfaces.

An argument could be made that the `Rollbar\Senders\AgentSender` does not require the access token. For now, I am going to stick with the conclusion from #405, and leave it the way it is.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix #469

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -291,13 +291,14 @@ class Config
         $this->setLogPayloadLogger($config);
         $this->setVerbose($config);
         $this->setVerboseLogger($config);
+        // The sender must be set before the access token, so we know if it is required.
+        $this->setSender($config);
         $this->setAccessToken($config);
         $this->setDataBuilder($config);
         $this->setTransformer($config);
         $this->setMinimumLevel($config);
         $this->setReportSuppressed($config);
         $this->setFilters($config);
-        $this->setSender($config);
         $this->setScrubber($config);
         $this->setBatched($config);
         $this->setBatchSize($config);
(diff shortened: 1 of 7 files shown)


waltjones (approved):
:+1:

PR #601, opened 2023-02-15
Updates for Laravel10 and Monolog 3

michaelscola:
## Description of the change

Updates required for Laravel 10

## Type of change

- [X] Maintenance
- [ ] New release

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,10 +31,10 @@
   },
 
   "require": {
-    "php": ">=8.0.0 <9.0",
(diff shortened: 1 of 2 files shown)


danielmorell:
@michaelscola I really don't want to drop support for PHP 8.0. I have created an alternative solution to add support for Monolog v3 in #602. If you have time to look at it and let me know what you think. I would appreciate it.

michaelscola:
A better solution was proposed.

PR #604, opened 2023-02-20
Fixed PHP 8.2 deprecated dynamic property creation warning

### LINKED ISSUE #590: PHP 8.2 Compatibility
With the release of PHP 8.2 there were a number of things that were [deprecated](https://www.php.net/manual/en/migration82.deprecated.php). Fortunately, only one appears to impact the Rollbar PHP package. It is **Deprecate dynamic properties**.

We use dynamic properties to keep track of errors and exceptions that were explicitly passed to the Rollbar or "caught" by the handlers via the `isUncaught` property.

This can easily be fixed in both our error handler and fatal error handler since we use our own `ErrorWrapper` class, and we can just declare the property in it. For the exception handler we will need to think through how we handle it.

danielmorell:
This PR fixes the PHP 8.2 deprecated dynamic property creation warning issue #590. 

## Description of the change

This adds a new class `ExceptionWrapper` to the SDK. It is used by the `ExceptionHandler::handle()` and `Rollbar::logUncaught()` methods. Both functions need to inform the `RollbarLogger::report()` method that the `Exception` instance we are passing it is on that was caught by the Rollbar SDK. To do that we had been setting the `isUncaught` property on the `Exception` instance. However, the `Exception` class does not have an `isUncaught` property. This means we were dynamically creating the property; something PHP allows. However, [in PHP 8.2 dynamic property creation has been deprecated.](https://www.php.net/releases/8.2/en.php#deprecate_dynamic_properties)

`ExceptionWrapper` let's us wrap the exception so that it can be passed to the `RollbarLogger::log()` method and include an `isUncaught` property while keeping the `log()` method compliant with the PSR logging interface.

For test review purposes I pushed the test to the branch first. [You can see the failed test in our CI](https://github.com/rollbar/rollbar-php/actions/runs/4226642287/jobs/7340297286#step:7:72) (while the test logs are retained). Also, note, that in the latest commit to this branch the test passes.

This also fixes all the dynamic properties I could find in the test suite.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

- Fix #590

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/Rollbar.php
+++ b/src/Rollbar.php
@@ -249,13 +249,8 @@ class Rollbar
         if (is_null(self::$logger)) {
             return self::getNotInitializedResponse();
         }
-        $toLog->isUncaught = true;
-        try {
-            $result = self::$logger->report($level, $toLog, $context);
-        } finally {
-            unset($toLog->isUncaught);
-        }
-        return $result;
+        $wrapper = new ExceptionWrapper($toLog, isUncaught: true);
+        return self::$logger->report($level, $wrapper, $context);
     }
 
     /**
--- a/src/ErrorWrapper.php
+++ b/src/ErrorWrapper.php
@@ -6,7 +6,7 @@ class ErrorWrapper extends \Exception
 {
     private static $constName;
 
-    public $isUncaught;
+    public bool $isUncaught = false;
 
     private $utilities;
 
--- a/src/ExceptionWrapper.php
+++ b/src/ExceptionWrapper.php
@@ -1,0 +1,47 @@
+<?php declare(strict_types=1);
+
+namespace Rollbar;
+
+use Stringable;
+use Throwable;
+
+/**
+ * The exception wrapper class is used to annotate an exception with a caught/uncaught flag. This is used internally by
+ * our handlers when passing exceptions to the {@see RollbarLogger::log()} method, since it must conform to the PSR
+ * logger interface.
+ */
+class ExceptionWrapper implements Stringable
+{
+    /**
+     * Instantiates the exception wrapper.
+     *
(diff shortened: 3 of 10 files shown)


src/Rollbar.php
```diff
@@ -249,13 +249,8 @@ public static function logUncaught(string|Level $level, Throwable $toLog, array
...
         }
-        $toLog->isUncaught = true;
-        try {
-            $result = self::$logger->report($level, $toLog, $context);
-        } finally {
-            unset($toLog->isUncaught);
-        }
-        return $result;
+        $wrapper = new ExceptionWrapper($toLog, isUncaught: true);
+        return self::$logger->report($level, $wrapper, $context);
```
anler, line 253: Suggestion: 

Why not add another optional parameter to `report` that is `bool $uncaught = false`? I find it better because is only that method the one needing that type of information, besides I find it confusing that we pass a `string | Stringeable` type to `report` but it internally coerces it into `mixed` when passing it to `isUncaughtLogData`.

If we do this then wrapper types for the exceptions and `isUncaughtLogData` would no longer be needed. The problem is `isUncaughtLogData` is public 😅 but I'm failing to see why is it useful hence my desire to remove it altogether.

danielmorell: That is a really good question.

I was trying to keep `report()` as close to `log()` as possible. I had not considered adding an optional argument, but that would definitely work.

`ExceptionHandler` currently calls the `RollbarLogger::log()` method, but that could be updated to use `report()`, if we went with the optional argument.

Overall, I think that using the optional argument solution would probably reduce complexity. I like it! I'll work on that.

PR #606, opened 2023-02-22
Fixed PHP 8.2 deprecated dynamic property creation 

### LINKED ISSUE #590: PHP 8.2 Compatibility

danielmorell:
This PR is an alternative to #604 and fixes the PHP 8.2 deprecated dynamic property creation warning issue #590.

## Description of the change

This adds a new optional argument to the `RollbarLogger::report()` method, `bool $isUncaught = false` as proposed by @anler https://github.com/rollbar/rollbar-php/pull/604#discussion_r1112973262. I created this as a new branch to keep our git history cleaner and then cherry picked the relevant commits from #604.

Because our internal `report()` method is not defined by the PSR log interface we are free to differ from the standard logging methods (as we already have done by returning a `Response` instance). Adding the argument removes the need to dynamically declare the `isUncaught` property as we had before.

A few things to note:

- The only reason this is considered a breaking change is because we are removing the public `RollbarLogger::isUncaughtLogData()` method since it is no longer needed.
- At every instance where we pass in the `$isUncaught` argument I use a named argument, so it is explicitly clear what the `true` is for.
- I did not add the `$isUncaught` argument to the `Rollbar::report()` method since we call `RollbarLogger::report()` directly, and the purpose of the argument is primarily for internal use. However, this does mean that the two `report()` methods are no longer the same. I am curious what the thoughts are on this.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fixes #590 
- Alternate to #604

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/ErrorWrapper.php
+++ b/src/ErrorWrapper.php
@@ -6,8 +6,6 @@ class ErrorWrapper extends \Exception
 {
     private static $constName;
 
-    public $isUncaught;
-
     private $utilities;
 
     private static function getConstName($const)
--- a/src/Handlers/ErrorHandler.php
+++ b/src/Handlers/ErrorHandler.php
@@ -42,9 +42,7 @@ class ErrorHandler extends AbstractHandler
         $exception = $this->logger()->
                             getDataBuilder()->
                             generateErrorWrapper($errno, $errstr, $errfile, $errline);
-        $exception->isUncaught = true;
-        $this->logger()->log(Level::ERROR, $exception, array());
-        unset($exception->isUncaught);
+        $this->logger()->report(Level::ERROR, $exception, isUncaught: true);
 
         return false;
     }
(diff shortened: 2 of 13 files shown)


anler (approved):
🎉

Issue #617, opened 2023-04-12
RollbarLogger::log() incompatible with Psr\Log\LoggerInterface::log()  v1.0


lukastaferner:
Dear Rollbar-Team,
Installing the package with psr/log:^1 produces the following error:
Fatal error: Declaration of Rollbar\RollbarLogger::log($level, Stringable|string $message, array $context = []): void must be compatible with Psr\Log\LoggerInterface::log($level, $message, array $context = []) in /srv/apache2/keepersport/vendor/rollbar/rollbar/src/RollbarLogger.php on line 194

Thank you!

Chris53897:
(I am not from the rollbar team)
I am not sure if it is possible to support psr/log 1-3. I did two PR and i suggest to remove support for version 1.

PR #618, opened 2023-04-16
Feature/chore add test --prefer-lowest

Chris53897:
Helps to reproduce for example https://github.com/rollbar/rollbar-php/issues/617
See https://github.com/Chris53897/rollbar-php/actions/runs/4713253520


danielmorell:
@Chris53897 Thank you for your work on this. Can you squash your commits into a single commit, so we can keep our change history clean?

danielmorell:
Closing in favor of #624

PR #619, opened 2023-04-16
chore: fix tests for performance test-suite

Chris53897:
this test suite does not run on CI (on purpose)
I just come across this, after i usally just run phpunit and do not select a testsuite.


danielmorell (approved):
This looks great! Thank you @Chris53897!

PR #621, opened 2023-04-16
Feat: allow phpunit 10

Chris53897:
- will set a default Testsuite
- use `static` data generators
- fix bug in ci for XDebug-Mode (see https://github.com/Chris53897/rollbar-php/actions/runs/4714282315/jobs/8360543577 and https://github.com/shivammathur/setup-php#ini-values-optional)
- `setMethods` => `onlyMethods`
- Exclude files and directories
- small compatibiliy layer for `assertObjectHasProperty`


danielmorell (approved):
@Chris53897 This looks great 🔥 

I reverted the drop of support for psr/log v1 since that would be a breaking change for some users, and fixed some of the merge conflicts.

I appreciate the work you put into this! It looks great!

PR #624, opened 2023-04-21
chore: add ci tests for --prefer-lowest

Chris53897:
Will replace https://github.com/rollbar/rollbar-php/pull/618
IDE would not allow squash it. A new PR was just faster for me


danielmorell (approved):
This looks great! Thank you @Chris53897!

I think I have a fix for the failing test on the lowest run. I'll submit that in a separate PR.

PR #635, opened 2024-03-01
Fixed #480 IP could not resolve IPv6

### LINKED ISSUE #480: Usage with IPv6 only (without IPv4)
On a server without an IPv4 address (but an IPv6), no message will get delivered. 
#35 did make sure of that. idk why this line should be set by default, but it at least has to be optional.

danielmorell:
## Description of the change

This fixes issue #480. It adds an optional config parameter `ip_resolve` that can be given one of the `CURL_IPRESOLVE_*` constants. Unless this config parameter is included the SDK will function entirely as before resolving with IPv4. This is not a breaking change.

_Note: Unfortunately, the PHP curl extension does not allow the retrieval of the value of this option once set. There is also no clear and consistently testable side effect that could be tested without introducing flaky tests. I tried a couple different methods but was unable to get consistent results. Because of that no new tests have been written._

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix #480

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/Senders/CurlSender.php
+++ b/src/Senders/CurlSender.php
@@ -26,6 +26,7 @@ class CurlSender implements SenderInterface
     private $maxBatchRequests = 75;
     private $batchRequests = array();
     private $inflightRequests = array();
+    private $ipResolve = CURL_IPRESOLVE_V4;
 
     public function __construct($opts)
     {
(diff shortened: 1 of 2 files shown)


src/Senders/CurlSender.php
```diff
@@ -54,6 +55,10 @@ public function __construct($opts)
         if (array_key_exists('ca_cert_path', $opts)) {
             $this->caCertPath = $opts['ca_cert_path'];
         }
+        if (array_key_exists('ip_resolve', $opts)
```
alif, line 58: Can we add testing for this scenario?

danielmorell: I added some tests to ensure that the value is set. I usually try to avoid testing private properties since they are generally assumed to be safe to refactor. So, I added a comment to the test in case it fails in the future because of some sort of refactoring.

alif (dismissed):

src/Senders/CurlSender.php
alif: I approved the PR but it looks like tests are failing.

alif (approved):

PR #663, opened 2025-10-04
Fixed #463 PII can be leaked if nested query string keys are out of order

### LINKED ISSUE #463: Scrubbing bug causing PII leaks in Rollbar PHP library
Feedback from one of the users:

> While investigating the leak of some PII information from our systems to Rollbar I identified a pretty serious bug in your PHP library (https://github.com/rollbar/rollbar-php).
> 
> Because of how the scrubbing feature is implemented, form submissions that cause exceptions and subsequently trigger Rollbar reporting can be excluded from the usual scrubbing process under certain circumstances.
>
> Technical explanation:
> In src/Scrubber.php on line 81-82 you are serializing in this case the body of a request containing form fields into an array. Usually this is no problem. However, when you have form arrays and they appear out of order it can cause line 81 to rearrange the hierarchy of the request (because of the array html fields), which causes the line 82 condition to fail which then means scrubbing is skipped on that field.
> 
> Scrubbing works:
> `login[username]=test@test.com&login[password]=secret`
>
> Scrubbing fails:
> `login[username]=test@test.com&unrelatedField=123&login[password]=secret`
>
> When scrubbing fails, passwords and other sensitive PII information are sent to Rollbar exempt from the usual scrubbing process.
>
> We're implementing mitigations on our end to prevent this happening in the short term but is there any chance this bug could be fixed?

danielmorell:
## Description of the change

This PR resolves an issue where out of order nested query string values could fail to be properly scrubbed.

The reason was we were parsing then rebuilding the query string to see if they matched. If they did we would decode and scrub the string as if it was a query string.

However, this caused issues because the nested keys would be order together, when the encoded string did not strictly require that. This solves that problem by partially decoding the string and comparing that with the decoded and rebuilt query string with a sorted key order.

As part of this change we also no longer encode all query strings that we send in the payload in the RFC3986 percent encoding sytanx. Instead we send the query string with the more human readable non-encoded characters.

## Type of change

## Related issues

- Fixed #463 

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+* Query strings no longer forced strings to have percent encoding.
+### Fixed 
+* Fixed #463 PII can be leaked if nested query string keys are out of order.
 
 ## [4.1.4] - 2025-09-26
 ### Fixed
--- a/src/Scrubber.php
+++ b/src/Scrubber.php
@@ -124,25 +124,47 @@ class Scrubber implements ScrubberInterface
     {
         if (is_array($data)) {
             // scrub arrays
-            $data = $this->scrubArray($data, $fields, $replacement, $path);
-        } elseif (is_string($data)) {
+            return $this->scrubArray($data, $fields, $replacement, $path);
+        }
+        if (is_string($data)) {
             // scrub URLs and query strings
             $query = parse_url($data, PHP_URL_QUERY);
             if ($query) {
-                $data = str_replace(
+                return str_replace(
                     $query,
                     $this->scrubQueryString($query, $fields),
                     $data
                 );
-            } else {
-                // PHP reports warning if parse_str() detects more than max_input_vars items.
-                @parse_str($data, $parsedData);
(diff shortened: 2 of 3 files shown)


brianr (commented):
@danielmorell This part I'm not understanding:

> As part of this change we also no longer encode all query strings that we send in the payload in the RFC3986 percent encoding sytanx. Instead we send the query string with the more human readable non-encoded characters.

Doesn't that mean that other characters that previously would have been encoded (and need to be encoded), will no longer be encoded?

Codex's review noted something similar:

```
- [P0] Keep percent-encoding when rebuilding query — src/Scrubber.php:233-236
  Calling urldecode() here means any percent-encoded characters (for example %26 representing &, %2B
representing +, etc.) are decoded before we hand the string back. Because we re-insert the string
without re-encoding, a query like token=abc%26def now becomes token=abc&def, which splits the value
into a brand-new parameter and permanently corrupts the payload we send. This regresses behaviour
for every query string that contains encoded separators, so we need to drop the urldecode() (and
accompanying str_replace) and keep the percent-encoding intact.
```

danielmorell:
We now check the original string to see if it contains any valid percent encoded hexidcimal characters. If, it does we will output a string with all characters that should be percent encoded in a URL percent encoded.

However, if the string does not contain any percent encoded characters. Even if we scrub the string as query string we will ensure we don't percent encode any characters that could be percent encoded.

This helps keep scrubbed content as close as possible to the same format we recieved it in.

brianr (approved):

